PR #15, opened 2017-10-20
Search for style.css in `/usr/bin/` too

GabMus:
By the way, I think that your /usr(/local)/bin/bottles directory should be /usr(/local)/share/bottles instead.


mirkobrombin:
Because of the need for a postinstall in the debian package, the package copy the files to a directory and move it through the setup.py file, this causes multiple directories because they are not all accessible by setup.py.
Thanks to PR!

PR #83, opened 2021-01-07
fix: The AppData files now go to /usr/share/metainfo

tim77:
# Description
https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#sect-Quickstart-DesktopApps

## Type of change

# How Has This Been Tested?
By appstream-glib tool https://github.com/hughsie/appstream-glib. It's available on Flathub and Linux distros.


mirkobrombin:
Thanks for your contribution. Before accepting your PR I have to make sure that it doesn't break the AppImage build.

mirkobrombin:
Nothing breaks, thanks again!

PR #272, opened 2021-05-14
style: Replace .btn_radius by .circular and .image-button

Kekun:
Let's use the standard style classes here. This also fixes the outline's
radius.

### DIFF
--- a/src/ui/breeze.css
+++ b/src/ui/breeze.css
@@ -13,8 +13,3 @@ button {
   border-radius: 0;
   box-shadow: none;
 }
-
-button.btn_radius {
-  border-radius: 100%;
-  padding: 10px;
-}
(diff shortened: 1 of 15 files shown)


mirkobrombin:
I've been looking for class documentation for a long time, do you have a References link?

Kekun:
Sadly there isn't a documentation of the Adwaita stylesheet AFAIK, but some widgets document classes affecting them in their documentation in a "CSS node" section, e.g. .circular is documented here: https://developer.gnome.org/gtk3/stable/GtkButton.html

That's probably something we should improve with libadwaita.

mirkobrombin:
> Sadly there isn't a documentation of the Adwaita stylesheet AFAIK, but some widgets document classes affecting them in their documentation in a "CSS node" section, e.g. .circular is documented here: https://developer.gnome.org/gtk3/stable/GtkButton.html
> 
> That's probably something we should improve with libadwaita.

Yes, this would definitely be a huge improvement. Thanks so much!

Issue #432, opened 2021-08-05
[BUG] Opening a .exe does nothing


sonnyp:
Bottles is fantastic thank you!

Possibly related #132 

**Describe the bug**

Opening an .exe with Bottles does nothing.

**Installation**
 - Method: flathub
 - Version 2021.7.28-treviso
 
**To Reproduce**
1. Download https://www.escapemotions.com/products/rebelle/download/Rebelle_4_Demo_64bit_v4.1.2_Windows.exe
2. Open with Bottles (suggested default action since that's my only .exe app handler)

**Expected behavior**

Someting happens, create a new bottle for the exe?

**Actual behavior**

(This is probably missing from the issue template)

Bottles opens, the progress cursor shows for a few seconds and nothing happens.

**Desktop (please complete the following information):**
 - OS: Fedora
 - Version 34

mirkobrombin:
Hi, thanks for being so explanatory. I think it's a regression in the CLI, can you try to start it from the terminal?

```bash
flatpak run com.usebottles.bottles -e path.exe
```

sonnyp:
Same.

```
sonny@porygon ~> flatpak run com.usebottles.bottles -e Downloads/Rebelle_4_Demo_64bit_v4.1.2_Windows.exe
Gtk-Message: 12:21:36.672: Failed to load module "canberra-gtk-module"
Gtk-Message: 12:21:36.672: Failed to load module "pk-gtk-module"
2021-08-05 12:21:36 INFO: Dxvk found: [dxvk-1.9.1]
2021-08-05 12:21:36 INFO: Vkd3d found: [vkd3d-v2.4]
2021-08-05 12:21:36 INFO: Runners found: [vaniglia-6.13]
2021-08-05 12:21:37 INFO: Bottles Started!
TypeError: Must be number, not NoneType
```

But this time I can see what I imagine the UI to create a Bottle appear and then replaced immediately with the welcome screen.

I noticed a couple of related issues around flatpak permissions and the .desktop file, I'll send a PR once opening from the CLI works.

sonnyp:
Actually the TypeError and the UI switch happens without an argument as well

```
sonny@porygon ~> flatpak run com.usebottles.bottles
Gtk-Message: 12:28:46.668: Failed to load module "canberra-gtk-module"
Gtk-Message: 12:28:46.668: Failed to load module "pk-gtk-module"
2021-08-05 12:28:46 INFO: Dxvk found: [dxvk-1.9.1]
[...]

Issue #413, opened 2021-07-30
[Flatpak] Reduce filesystem permissions 


jannuary:
(Moved from [com.usebottles.bottles#120](https://github.com/flathub/com.usebottles.bottles/issues/120))

Currently, Bottles has a whopping 12 different filesystem permissions. Now that we use portals, it wouldn't hurt to go through different permission and reduce ones that aren't needed.

sonnyp:
With https://github.com/bottlesdevs/Bottles/pull/436 - all filesystem permissions that are related to letting a user open an arbitrary file can be removed. As far as Bottles is concerned.

Regarding Windows app that open/save files;

I see 3 options

1. Make Bottles request rw permission on `~`
2. Remove all filesystem permissions and instruct users to use [Flatseal](https://github.com/tchx84/Flatseal) if they need a permission for a Windows app
3. Let users pick folders (via the portal) which are exposed to a Bottle via a symbolic link in the wine prefix (is that even possible?)

WDYT?

mirkobrombin:
> With #436 - all filesystem permissions that are related to letting a user open an arbitrary file can be removed. As far as Bottles is concerned.
> 
> Regarding Windows app that open/save files;
> 
> I see 3 options
> 
> 1. Make Bottles request rw permission on `~`
> 2. Remove all filesystem permissions and instruct users to use [Flatseal](https://github.com/tchx84/Flatseal) if they need a permission for a Windows app
> 3. Let users pick folders (via the portal) which are exposed to a Bottle via a symbolic link in the wine prefix (is that even possible?)
> 
> WDYT?

Surely the second and third points are the best. What I don't understand is how it is possible that in my tests, even removing the permissions I am able to browse and pick files from the system using wine explorer.

mirkobrombin:
![image](https://user-images.githubusercontent.com/9479903/128350902-7478eccf-eb6d-4cf3-8214-65e6627575f3.png)

Maybe the sandbox is disabled by GNOME Builder or flatpak-builder?

sonnyp:
> Maybe the sandbox is disabled by GNOME Builder or flatpak-builder?

Exactly

mirkobrombin:
> > Maybe the sandbox is disabled by GNOME Builder or flatpak-builder?
> 
> Exactly

This explains **A LOT** of things. It is time for me to start using more Flatpak packages.

sonnyp:
[...]

PR #436, opened 2021-08-05
Add file opening to desktop file

sonnyp:
Helps with #413 

# Description

1. This helps the desktop understand how file opening works with the app. See https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#exec-variables

2. In addition, and more importantly, this let flatpak users open any file with Bottles, no matter where they are and without fs permissions. Without this Right Click -> Open with Bottles cannot work .

```sh
# Build and install the flatpak
flatpak-builder --user --force-clean --install-deps-from=flathub --install flatpak com.usebottles.bottles.json
# Print the desktop file
cat ~/.local/share/flatpak/exports/share/applications/com.usebottles.bottles.desktop

# Before
Exec=/usr/bin/flatpak run --branch=stable --arch=x86_64 --command=bottles com.usebottles.bottles

# After
Exec=/usr/bin/flatpak run --branch=master --arch=x86_64 --command=bottles --file-forwarding com.usebottles.bottles @@ %f @@
```

Independently of this PR, In general you should avoid encouraging users from using Bottles flatpak from CLI as it breaks some assumptions, but if you must, it's important to instruct as such:

```
flatpak run --file-forwarding com.usebottles.bottles @@ path.exe @@
```

See `--file-forwarding` on https://docs.flatpak.org/en/latest/flatpak-command-reference.html

## Type of change

# How Has This Been Tested?

```
flatpak-builder --user --force-clean --install-deps-from=flathub --install flatpak com.usebottles.bottles.json
cat ~/.local/share/flatpak/exports/share/applications/com.usebottles.bottles.desktop
```

Unfortunately because of #432 I couldn't assert that right click -> Open with Bottles on a file which isn't readable by Bottles works. 

This is standard though and what every application that can open a file does. Happy to test once #432 is solved.

### DIFF
--- a/data/com.usebottles.bottles.desktop.in
+++ b/data/com.usebottles.bottles.desktop.in
@@ -2,7 +2,7 @@
 Name=Bottles
 Comment=Easily manage wineprefix
 Icon=com.usebottles.bottles
-Exec=bottles
+Exec=bottles %f
 Terminal=false
 Type=Application
 Categories=Utility;


mirkobrombin:
> I can confirm that Right Click -> Open with Bottles works with this even if the file isn't readable by Bottles (at `~/file.exe` for example)
> 
> (If you already have a Bottle as explained in [#432 (comment)](https://github.com/bottlesdevs/Bottles/issues/432#issuecomment-893363301))

Thanks. Can I merge or need more testing?

PR #444, opened 2021-08-07
Use HdyActionRow for environment rows

### LINKED ISSUE #443: [BUG] The use of color is confusing
First of all, I really like the UI refresh, kudos to the designers.

I have an issue with the use of colors for Gaming/Software/Custom

This blue is Adwaita/GNOME color for primary actions
This red is Adwaita/GNOME color for destruction actions
This grey is Adwaita/GNOME colour for insensitive (disabled) actions

IMHO it's very confusing

Screenshot from 2021-08-07 16-06-49

Regardless of the colors being used, it's also unusual for GNOME applications to make use of such strong color accents.

I think the symbolic icons for each category is enough and I don't really see the need for the colors. They don't appear to be re-used (for this purpose) anywhere else.

If you agree I'm happy to send a PR to give a try or to any other proposal.

sonnyp:
* Fixes #443
* removes unused styes
* turn environment row/entry into HdyActionRow to follow GNOME HIG and simplify code
* fixes the margin between the label and the list
* always show subtitles/descriptions

Screenshot from 2021-08-08 01-27-13


src/ui/style.css
```diff
@@ -119,22 +85,8 @@
...
-
-.environment-entry.software {
-  background: #1C71D8;
-  border: 1px solid rgba(0, 0, 0, 0.2);
-}
-
-.environment-entry.custom {
-  background: #9A9996;
   border: 1px solid rgba(0, 0, 0, 0.2);
+  color: black
```
sonnyp, line 89: for some reason it's white by default (couldn't figure out why)

setting this to `color: black` breaks dark mode, any idea what's the best way to set this conditionally based on dark mode being preferred or not?

I don't think GTK supports `prefers-color-scheme` but I didn't try TBH

mirkobrombin:
So fast :D this may be The solution to the problem. I let the designers to take this decision.

src/ui/style.css
mirkobrombin: > for some reason it's white by default (couldn't figure out why)
> 
> setting this to `color: black` breaks dark mode, any idea what's the best way to set this conditionally based on dark mode being preferred or not?
> 
> I don't think GTK supports `prefers-color-scheme` but I didn't try TBH

prefers-color-scheme is not supported :S maybe we can offer 2 different styles and then inject when user switch the dark mode

sonnyp:
An other option would be to do a radio boxed list 

See "Homepage" section on the screenshot here:

https://developer.gnome.org/hig/patterns/containers/boxed-lists.html

It's more worked that I'm no necessarily interested in doing right now so perhaps in the future?

src/ui/style.css
mirkobrombin: maybe it is white from the parent

mirkobrombin:
> An other option would be to do a radio boxed list
> 
> See "Homepage" section on the screenshot here:
> 
> https://developer.gnome.org/hig/patterns/containers/boxed-lists.html
> 
> It's more worked that I'm no necessarily interested in doing right now so perhaps in the future?

Not a fan but we are trying to get closer to GNOME HIG

alvarlagerlof:
> An other option would be to do a radio boxed list 
> 
> See "Homepage" section on the screenshot here:
> 
> https://developer.gnome.org/hig/patterns/containers/boxed-lists.html
> 
> It's more worked that I'm no necessarily interested in doing right now but perhaps worth considering for the future?

I like the radio option if the current icons also move to the right. Otherwise I think we at least need a white background. Just the outline is too bare.

src/ui/style.css
sonnyp: Thanks - I'll look into it once we have a go from designers.

sonnyp:
>  Otherwise I think we at least need a white background. Just the outline is too bare.

Like this ?

Screenshot from 2021-08-07 19-46-27

mirkobrombin:
> > Otherwise I think we at least need a white background. Just the outline is too bare.
> 
> Like this ?
> 
> Screenshot from 2021-08-07 19-46-27

Looks more evident, I like it

sonnyp:
Ok turns out I had to make bigger changes so I'm fine implementing the radio solution

Just let me know of the 3 which you prefer @alvarlagerlof as a reminder

1. checkmark https://github.com/bottlesdevs/Bottles/pull/444#issuecomment-894684661
2. radio https://developer.gnome.org/hig/patterns/containers/boxed-lists.html
3. this:

Screenshot from 2021-08-07 20-33-30

src/ui/environment-entry.ui
```diff
@@ -2,101 +2,9 @@
...
-        <style>
-          <class name="selection-icon"/>
-        </style>
-      </object>
-      <packing>
-        <property name="expand">False</property>
-        <property name="fill">True</property>
-        <property name="position">2</property>
-      </packing>
-    </child>
```
sonnyp, line 99: @mirkobrombin this PR is quite a trim, are you ok with it?

Also, this is my first time doing Python GTK, I don't think we need this `.ui` anymore, how can I make `BottlesEnvironmentEntry` extend Handy.ActionRow without this template?

mirkobrombin: > @mirkobrombin this PR is quite a trim, are you ok with it?
> 
> Also, this is my first time doing Python GTK, I don't think we need this `.ui` anymore, how can I make `BottlesEnvironmentEntry` extend `Handy.ActionRow` without a template?
> 
> EDIT: nevermind, easy peasy

It's fine. This is also my first project using .ui with gtk

sonnyp (commented):
I updated the description.

Ready for review :rocket: 

EDIT: nevermind down below - I figured it out - `.content` class was missing. I don't remember needing it before for ListBox and it's not documented, if you know why it's needed I'd be happy to know.

It's almost done but I can't figure out why this list and its rows don't have borders and separators.

According to the GTK inspector, Adwaita.css rules do not apply to this list like it does for the one below but I can't find the reason or any difference between the 2 for that matter.

Any ideas?

Screenshot from 2021-08-08 00-59-21

mirkobrombin:
Not everything is documented in GTK. Btw great job! I want to test to understand the borders problem

jannuary:
Aside from 32px icons, LGTM

image

mirkobrombin:
> Aside from 32px icons, LGTM
> 
> image

Unfortunately according to the documentation, the icons cannot be resized D:

mirkobrombin:
image

Perhaps the color should also be removed from the tags, it could be confusing. Maybe using the same icons?

jannuary:
Do we need to display the bottle type this prominently in the list in the first place? The list is already quite visually heavy, so I'd prefer just removing these

mirkobrombin:
> Do we need to display the bottle type this prominently in the list in the first place? The list is already quite visually heavy, so I'd prefer just removing these

It depends on how many bottles you have on the list. In the tests I reached 20 and having a colored reference (or an icon) helps me a lot to filter them at least by type

mirkobrombin:
image

sonnyp:
The colors in the tags are unrelated.

I would appreciate if we could merge this instead of highjacking it.

mirkobrombin:
In my opinion they are related problems but ok. I'm fine with the merge

sonnyp:
> Do you know if it is possible somehow?

No idea but I would suggest to stick with libhandy defaults.

At least until the port to GTK4/libadwaita.

sonnyp:
Regarding the colors in the tags, I like them, Nautilus also has color tags and I don't think they pose any user experience problem.

mirkobrombin:
> Regarding the colors in the tags, I like them, Nautilus also has color tags and I don't think they pose any user experience problem.
> 
> its a common pattern

The point is that users see these colors after the bottle has been created. For the same problem, seeing the red background they may think it is a "problem" because the explanation of the colors is missing.

sonnyp:
It's not the same red as the one for destructive actions. The tag widget is also clearly not actionable.
Colored tags are a common pattern and I think it is clear to the users the color is associated with the word on it.

IMHO you shouldn't use those icons without labels because it's not clear what they are/mean. Even after creating.

I would suggest to leave as is - or - reconsider if these tags add any value in this list.

PR #460, opened 2021-08-11
Remove flatpak filesystem permissions

### LINKED ISSUE #413: [Flatpak] Reduce filesystem permissions

sonnyp:
Fixes #413


com.usebottles.bottles.elementary.json
```diff
@@ -5,31 +5,15 @@
...
         "--allow=multiarch",
         "--share=network",
         "--share=ipc",
         "--socket=wayland",
         "--socket=fallback-x11",
         "--socket=pulseaudio",
-        "--filesystem=~/Games:ro",
-        "--filesystem=~/wineprefixes:ro",
-        "--filesystem=~/.Bottles:ro",
-        "--filesystem=~/.local/share/bottles:ro",
```
sonnyp, line 18: @mirkobrombin the `BottlesFlatpakMigration` does not appear to be used, `show_flatpak_migration_view` is never called

Should I remove it?

mirkobrombin: BottlesFlatpakMigration is no more needed and can be safely removed. It's ok for the section in the readme

mirkobrombin (commented):
image
the only "problem" I found is with wine not being able to open or save files from/in any path due to the new permissions

mirkobrombin:
I think the second point is the better choice but should be implemented only for flatpak users as other packages has not these restrictions: https://github.com/bottlesdevs/Bottles/issues/413#issuecomment-893419654

sonnyp:
@mirkobrombin so you mean

> Remove all filesystem permissions and instruct users to use Flatseal if they need a permission for a Windows app

How do you feel about instructions in the README.md of the Flathub Bottles repository for now?

mirkobrombin:
That may be fine but most users won't reach that repository. Perhaps it is better to at least allow access to the Downloads folder in read mode. What do you think?

mirkobrombin:
I don't like the idea of ​​displaying the home but I'm afraid no user will appreciate this change and maybe it's too difficult to find that information

mirkobrombin:
Maybe the third option is the best 
> Let users pick folders (via the portal) which are exposed to a Bottle via a symbolic link in the wine prefix (is that even possible?)

With a simple tutorial the user will be introduced to the new feature.

sonnyp:
Nah, either we say "sandboxed" and we provide a way for users to punch holes at will. Or we say "Bottles needs fs" in which case we might as well leave gvfs, `/run/media` and so on.

I'm strongly in favor of the first even if it's more work. If users don't care about sandboxing they can use AppImage, Snap, distribution package, ...

What about we show a "Give filesystem access" button in the bottle to flatpak users that opens the documentation on how to for now?

Later (no promises) I would like to investigate into using the documen portal for this. There is a persistent option https://github.com/flatpak/xdg-desktop-portal/blob/89d2197002f164d02d891c530dcaa2808f27f593/data/org.freedesktop.portal.Documents.xml#L73

Related: https://github.com/flatpak/xdg-desktop-portal/issues/133

mirkobrombin:
Yes, I think it's a good idea to add the button.

src/window.py
```diff
@@ -23,7 +23,7 @@
 gi.require_version('Handy', '1')
 gi.require_version('Notify', '0.7')
 
-from gi.repository import Gtk, Gio, Notify, Handy
+from gi.repository import Gtk, Gio, Notify, Handy, GLib
```
sonnyp, line 26:
```suggestion
from gi.repository import Gtk, Gio, Notify, Handy
```

mirkobrombin:
We should merge and open a new issue for button documentation. Is it ok?

mirkobrombin (approved):

PR #800, opened 2021-12-15
update Debian packaging

SwampRabbit:
# Description
Update Debian packaging

Fixes:
- switch to debhelper-compat
- add libglib2.0-dev to build depends
- add gettext to build depends
- add relevant links
- add min version gir1.2-handy-1 (>=1.5.0) because it fails to launch without it
- update Description

How Has This Been Tested?
Built successfully in pbuilder schroot build on Debian 11
Installed successfully on Debian 11 and MX-21 VirtualBox VMs (libhandy-1 1.5.0 from Debian Bookworm was needed though)
Launches, installed components through first time run, and tested creating a bottle


mirkobrombin:
Thanks!

PR #1000, opened 2022-01-31
Add missing return in __get_cwd method

asbytes:
# Description
Added missing return in `WineCommand.__get_cwd` method.
Some apps crash when process CWD is not set to exe directory on process launch.

## Type of change


mirkobrombin:
How did I miss this :| ? Thanks, nice found!

mirkobrombin:
Fortunately, we have time to distribute it before the first hotfix hit flathub

PR #1105, opened 2022-03-01
Use GNotification instead of libnotify

A6GibKm:
This is more sandbox friendly and does not require and extra library.

We also add X-GNOME-UsesNotifications=true so that GNOME Software and
similar can pick this piece of metainfo.

Ill test in a bit.

### DIFF
--- a/com.usebottles.bottles.dev.json
+++ b/com.usebottles.bottles.dev.json
@@ -14,7 +14,6 @@
     "--socket=pulseaudio",
     "--device=all",
     "--system-talk-name=org.freedesktop.UDisks2",
-    "--talk-name=org.freedesktop.Notifications",
     "--env=LD_LIBRARY_PATH=/app/lib:/app/lib32",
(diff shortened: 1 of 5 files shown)


mirkobrombin:
Good, I’ll test soon

A6GibKm:
It test it with a quick

```python
        self.send_notification(
            title="Bottles",
            text=_("You are offline, unable to download."),
            image="network-wireless-disabled-symbolic"
        )
```
in the `__init__` of the main window and it works. There is one caveat though, for notifications to work for an app, the app has to be installed so that the .desktop file is picked up, this is not related to the addition to the desktop file thats just for Software if I am not mistaken.

A6GibKm:
Yes, just pointing out that if for example you test a developer version (say flatpak or just running `python bottles`) without bottles installed it wont work.

mirkobrombin:
It makes sense, I personally build the flatpak on every test so I shouldn't have the problem

mirkobrombin (approved):

PR #1107, opened 2022-03-02
Support Dark Mode

A6GibKm:


A6GibKm:
So for comparison, here is how the pill tags now look

image

image

A6GibKm:
The approach for medals was to invert the colors which works surprisingly well

image

A6GibKm:
The only thing missing tweaks is `Dialog.__init__`, which should not only read the `gtk_style_managet_get_dark` property but also should update the styles on `notify::dark`, but reading the source I think that `html` and `log` are always False?.

mirkobrombin:
Very nice work!

> The only thing missing tweaks is `Dialog.__init__`, which should not only read the `gtk_style_managet_get_dark` property but also should update the styles on `notify::dark`, but reading the source I think that `html` and `log` are always False?.

Yup they are

mirkobrombin:
Should this PR moved to draft?

A6GibKm:
mmm let me move it to draft until I can test it, for some reason GNOME Builder is not building it. Btw the only change I did was fix a small error on the meson file and a rebase.

mirkobrombin:
> mmm let me move it to draft until I can test it, for some reason GNOME Builder is not building it. Btw the only change I did was fix a small error on the meson file and a rebase.

Ya, I see the action failing https://github.com/bottlesdevs/Bottles/runs/5397872635?check_suite_focus=true

A6GibKm:
Ok, managed to get it to build and test it and works :). CI is failing since the libhandy version in arch is 1.4, ill look into it, but one thing I have no clue how to deal with is the snap manifest.

francescomasala:
> Ok, managed to get it to build and test it and works :). CI is failing since the libhandy version in arch is 1.4, ill look into it, but one thing I have no clue how to deal with is the snap manifest.

Maybe we can try to use [this ](https://aur.archlinux.org/packages/libhandy-git)aur package but idk, i'll try to fix the CI

A6GibKm:
Yes, but I have no clue how to use the AUR

mirkobrombin:
> Ok, managed to get it to build and test it and works :). CI is failing since the libhandy version in arch is 1.4, ill look into it, but one thing I have no clue how to deal with is the snap manifest.

The snap is unpublished and broken, so just ignore it.

francescomasala:
> Yes, but I have no clue how to use the AUR

I'll do that, don't worry

mirkobrombin:
I didn't think about it before but third-party packages? 😨

There are some unofficial packages for Fedora, olensuse, Mx Linux, void, nix. Some of those distros can have no access to a more recent libhandy release.

A6GibKm:
> I didn't think about it before but third-party packages? fearful
> 
> There are some unofficial packages for Fedora, opensuse, Mx Linux, void, nix. Some of those distros can have no access to a more recent libhandy release.

Ideally we should just add libhandy as a subproject but CI is failing, even tough local builds work. The thing is that 1.5 is a developer release and 1.6 should be released soon, we can wait a bit for it.

mirkobrombin:
> > I didn't think about it before but third-party packages? fearful
> > There are some unofficial packages for Fedora, opensuse, Mx Linux, void, nix. Some of those distros can have no access to a more recent libhandy release.
> 
> Ideally we should just add libhandy as a subproject but CI is failing, even tough local builds work. The thing is that 1.5 is a developer release and 1.6 should be released soon, we can wait a bit for it.

I'll check the CI

mirkobrombin:
The problem with makepkg was a lack of permissions. Luckily, there's a new one: https://github.com/bottlesdevs/Bottles/runs/5410844497?check_suite_focus=true#step:6:372

A6GibKm:
We had a similar issue to that one on libhandy CI using fedora 35....

mirkobrombin:
Nice! Thanks again for your contribution. I see you've already updated with the recent changes too

mirkobrombin (approved):

PR #1225, opened 2022-03-20
style: use proper naming for v1

cat-master21:
# Description
The name com.github.mirkobrombin.bottles doesn't work well and breaks in some ways.

Fixes #(issue)
Fixes some issues lately with naming.

## Type of change


cat-master21:
Realized this is only needed for some files on /debian.

PR #1226, opened 2022-03-20
style: change deb naming for v1

cat-master21:
# Description
Using the longer deb name doesn't work out well in some ways and plus the name for bottles on the master branch is com.usebottles.bottles anyway which works out much better for Debian packaging.

Fixes #(issue)
Fixes installing on Debian in some ways.

## Type of change


cat-master21:
Never mind the name isn't an issue but it's still annoying.

PR #1242, opened 2022-03-27
Add some small improvements

psydvl:
# Description
Just some little improvements

## Type of change
- [x] Some non-breaking improvements


mirkobrombin:
Thanks! Is this ready to merge?

psydvl:
It's ready to merge, however, I found that there are alternative things used at the same time 
Like path with f-string and path with os.path.join 
I can replace all f-string with os.path.join 
Should I?

mirkobrombin:
I was already thinking about migrating to path join, I think it's better that way

PR #1299, opened 2022-04-11
Fix/appimage package

### LINKED ISSUE #1272: Bottles AppImage
cc @azubieta (GitHub doesn’t let me to assign the issue)

Bottles’ Appimage has been deprecated due to several issues. Let me make a small summary:

- packages like p7zip and cabextract are not always found even if they are present (in silverblue yes but Ubuntu and Arch no)
- WINE is unable to find winbind and 32bit dependencies (this may be related to the first point, i.e. that what is installed is not detectable by Bottles)
- Caffe (the default runner in Bottles) requires glibc 2.32 or higher
- for the first releases we used a custom script that did not include dependencies https://github.com/bottlesdevs/Bottles/blob/master/appimage-build.sh
- we switched to Appimage builder, this is the recipe which has the above problems https://github.com/bottlesdevs/Bottles/blob/master/AppImageBuilder.yml

Here is the list of necessary dependencies https://github.com/bottlesdevs/Bottles/wiki/Packaging#dependencies

azubieta:
# Description
Fix AppImage package using appimage-builder. Notice that AppImage were not able to share their runtime (libc, et. al) with external applications until now. In this recipe we are using the very latest appimage-builder which includes support for external modules.

External modules are defined by setting their paths using the environment variable `APPDIR_MODULE_DIR`. Additionally the modules needs to be patched after installation so they can use the AppImage runtime. Use the following script to patch the modules https://gist.github.com/azubieta/202c85f1f5a9e5c6be5cb066fcb2d5e2

**Additional changes need to be made in the Bottles module installation to add the patch mentioned above**

Fixes #(issue)

## Type of change

# How Has This Been Tested?

- Create the build environment described by the Dockerfile
- Build the appimage inside the docker file using `appimage-builder --skip-test`

Closes #1272

### DIFF
--- a/.github/workflows/build_pkgs_unstable.yml
+++ b/.github/workflows/build_pkgs_unstable.yml
@@ -58,7 +58,7 @@ jobs:
         run: |
           mv Bottles-*-x86_64.AppImage Bottles-devel-x86_64.AppImage
           mv Bottles-*-x86_64.AppImage.zsync Bottles-devel-x86_64.AppImage.zsync
-          mv ../com.usebottles.bottles*.deb com.usebottles.bottles-unstable.deb
+          # mv ../com.usebottles.bottles*.deb com.usebottles.bottles-unstable.deb
 
       - name: Upload AppImage to Artifacts
         uses: actions/upload-artifact@v2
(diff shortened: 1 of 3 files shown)


mirkobrombin:
I've restored the old github workflow, should be triggered on merge

mirkobrombin:
I've restored the old github workflow, should be triggered on merge

Issue #1325, opened 2022-04-15
[Bug]: Desktop file does not execute exe if the bottle has a space in its name


emiltang:
## Observed Behavior
When i click a desktop icon with a bottle with a space in its name it only opens the main page bottles

## Expected Behavior
Clicking the icon should run the exe in a bottle

### To Reproduce

1. Create a new bottle with a space in the name
2. Install a program
3. Create a desktop icon
4. Click icon

### Package

From source

### Distribution

Arch

### Health Check

```shell
Display:
    X.org: true
    X.org (port): :0
    Wayland: true
Graphics:
    vendors:
        intel:
            vendor: intel
            envs:
                DRI_PRIME: '1'
            icd: /usr/share/vulkan/icd.d/intel_icd.x86_64.json:/usr/share/vulkan/icd.d/intel_icd.i686.json
    prime:
        integrated: null
        discrete: null
Kernel:
    Type: Linux
    Version: 5.17.3-arch1-1
Distro:
    Name: Arch Linux
    Version: Unknown
Tools:
    cabextract: true
    p7zip: true
    patool: true
    glibc_min: '2.35'
Bottles_envs: null
```

### Display Protocol

Wayland

### Display Backend used by Bottles

Wayland

### Additional context

Pull request incoming in sec

mirkobrombin:
This could actually be a problem. Thanks for reporting

PR #1393, opened 2022-04-30
Change bottle-cli command to use Path

### LINKED ISSUE #1392: [Bug]: Desktop entry doesnt work if bottles name is changed
### Describe the bug

When a bottle is created it gets a name and a path. The name is the thing that;s displayed in the UI, while the path i the name of the wine prefix directory.

When a bottle is renamed only the names is changed. This is fine in my opinion.

The issue is that the desktop entry creator function assumes that the name and path is always the same. This is most defiantly not the case.

### To Reproduce

1. Create a bottle
2. Install an application
3. Change bottle name
4. Create a desktop entry
5. Click new desktop entry in the DE menu

## Expected behavior:
The wine application is launched.

## Observed behavior
Nothing happens

### Package

From source

### Distribution

Arch

### Health Check

```shell
Display:
    X.org: true
    X.org (port): :1
    Wayland: false
Graphics:
    vendors:
        nvidia:
            vendor: nvidia
            envs:
                __NV_PRIME_RENDER_OFFLOAD: '1'
                __GLX_VENDOR_LIBRARY_NAME: nvidia
                __VK_LAYER_NV_optimus: NVIDIA_only
            icd: /usr/share/vulkan/icd.d/nvidia_icd.json
    prime:
        integrated: null
        discrete: null
Kernel:
    Type: Linux
    Version: 5.17.4-arch1-1
Distro:
    Name: Arch Linux
    Version: Unknown
Tools:
    cabextract: true
    p7zip: true
    patool: true
    glibc_min: '2.35'
Bottles_envs: null
```

### Display Protocol

X11

X11

### Additional context

_No response_

emiltang:
# Description
There are many cases where the bottle name and path are not the same. Launching a bottle with CLI requires the bottle path not the bottles name. It only works in the case where they are equal.

Fixes #1392  #1325

## Type of change
- [ *] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?
*    Create a bottle
*    Install an application
*    Change bottle name
*    Create a desktop entry
*    Click new desktop entry in the DE menu
*    Check that application launches correctly

### DIFF
--- a/src/backend/utils/manager.py
+++ b/src/backend/utils/manager.py
@@ -181,7 +181,7 @@ class ManagerUtils:
         with open(desktop_file, "w") as f:
             f.write(f"[Desktop Entry]\n")
             f.write(f"Name={program.get('name')}\n")
-            f.write(f"Exec={cmd} run -p '{program.get('name')}' -b '{config.get('Name')}'\n")
+            f.write(f"Exec={cmd} run -p '{program.get('name')}' -b '{config.get('Path')}'\n")
             f.write(f"Type=Application\n")
             f.write(f"Terminal=false\n")
             f.write(f"Categories=Application;\n")


mirkobrombin:
Thanks for your contribution!

PR #1465, opened 2022-05-14
Formally adopt GNOME CoC

jannuary:
much overdue :p

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,126 @@
+# GNOME Code of Conduct
+
+Thank you for being a part of the GNOME project. We value your participation and want everyone to have an enjoyable and fulfilling experience. Accordingly, all participants are expected to follow this Code of Conduct, and to show respect, understanding, and consideration to one another. Thank you for helping make this a welcoming, friendly community for everyone.
+
+## Scope
+
+This Code of Conduct applies to all online GNOME community spaces, including, but not limited to:
+
+ * Issue tracking systems - bugzilla.gnome.org
+ * Documentation and tutorials - developer.gnome.org
+ * Code repositories - git.gnome.org and gitlab.gnome.org
+ * Mailing lists - mail.gnome.org
+ * Wikis - wiki.gnome.org
+ * Chat and forums - irc.gnome.org, discourse.gnome.org, GNOME Telegram channels, and GNOME groups and channels on Matrix.org (including bridges to GNOME IRC channels)
+ * Community spaces hosted on gnome.org infrastructure
+ * Any other channels or groups which exist in order to discuss GNOME project activities
+
+Communication channels and private conversations that are normally out of scope may be considered in scope if a GNOME participant is being stalked or harassed. Social media conversations may be considered in-scope if the incident occurred under a GNOME event hashtag, or when an official GNOME account on social media is tagged, or within any other discussion about GNOME. The GNOME Foundation reserves the right to take actions against behaviors that happen in any context, if they are deemed to be relevant to the GNOME project and its participants.
+
+All participants in GNOME online community spaces are subject to the Code of Conduct. This includes GNOME Foundation board members, corporate sponsors, and paid employees. This also includes volunteers, maintainers, leaders, contributors, contribution reviewers, issue reporters, GNOME users, and anyone participating in discussion in GNOME online spaces.
+
+## Reporting an Incident
+
+If you believe that someone is violating the Code of Conduct, or have
+any other concerns, please [contact the Code of Conduct committee](https://wiki.gnome.org/Foundation/CodeOfConduct/ReporterGuide).
+
+## Our Standards
+
+The GNOME online community is dedicated to providing a positive experience for everyone, regardless of:
+
+ * age
+ * body size
+ * caste
+ * citizenship
+ * disability
+ * education
+ * ethnicity
+ * familial status
+ * gender expression
+ * gender identity
+ * genetic information
+ * immigration status
+ * level of experience
+ * nationality
+ * personal appearance
+ * pregnancy
+ * race
+ * religion
+ * sex characteristics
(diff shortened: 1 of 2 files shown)


TheEvilSkeleton (approved):

mirkobrombin (approved):

Martinligabue (approved):
seems a nice thing

mirkobrombin:
Asked for review to main contributors (excluding translations)

Martinligabue (approved):

francescomasala (approved):

pietrodicaprio (approved):

sonnyp (approved):

alvarlagerlof (approved):

PsykeDady (approved):

comradekingu:
This is the reason I don't contribute to GNOME anymore.

@Martinligabue
I will agree it seems nice, with the greatest possible emphasis on "seems".
This is by design. The trajectory is the same, and it ends up the same.
Meanwhile it wastes time, and it produces disasters with no tangible benefit.

@jannuary 
What does "This also includes repeated subtle and/or indirect discrimination." actually mean, and why is the scope of everything extended to "private conversations"?
What exactly is "sexualized clothing", and where/how is the line drawn?
What makes sexual advances (ipso facto) inappropriate?
What sexual advances are possible if their only redeeming quality is that all of them should be successful?
If the only possible answer is to ask for the permission to flirt, is that how humans operate, anywhere, at any point in time?
What does "The GNOME community prioritizes marginalized people's safety over privileged people's comfort." mean?
Why are those at odds with each-other?
Where does one stop and end, and what system is this denominated in?
Are community members meant to keep an internal list of either group?

If the answer is to not care for what this document actually says, I still wonder why it formulated in such a way.
If it is impossible to tell, I think it opens up the potential for misinterpretation.

Further, why does "the committee" not act on:

>"Reverse"-isms, including "reverse racism," "reverse sexism," and "cisphobia"

>Criticizing racist, sexist, cissexist, or otherwise oppressive behavior or assumptions.

---

I will indulge in the created opportunity to criticise it.
Making the earlier caveat and then following it up with its own mutually exclusive point, twice, is a baffling exercise in logic.
The URL includes "#supporting-diversity", which is by its context removed from the meaning of that word.
There is exclusive discrimination at play here, assuming that irrational fear of heterosexual people doesn't exist, or doesn't matter.
It should be said the term "phobia" is misappropriated from actual phobias in the medical sense, but the use here hinges on the idea that an opposite phobia is real.
Moreover, one can deduce "racism" is also not meant to be all-encompassing, but rather selectively applied from an understanding of … what exactly? My take is that the tautology of adding the word "reverse" is indistinguishable from the original word and its meaning.
The irony is also not lost on adopting the word "sex" to not be inclusive in that same sentence.
At face value, "cissexist" has to be a subset of "sexism" pertaining only to sexism facing men or women, yet that has its own listing. Similarly to those terms, otherwise "transsexist" becomes possible, though that is omitted…
Again one wonders if doing it is somehow different based on not to whom it is done, or the group doing it…
When the group is assumed and the terms made up, it is all a game against someone with all the cards.

The group identifying "oppressive behavior" or even "assumptions" isn't defined, which I take to mean critique in all forms is OK so long as it finds a target in the form of pointing out those. I don't use Twitter, but from what I gather this is the current game played.

When the scope of racism stops being race, and sex starts to also include arbitrary understandings, problems are not solved, but rather created. One of which is that the supposed _equal treatment of all_ necessitates words that are not only counterfactual, but also one would assume counterproductive.
Not that this type of departure from objectivity has ever solved anything, but alas.

I support the Bottles maintainer in saying [this](https://github.com/bottlesdevs/Bottles/pull/661#issuecomment-952747810)
which is a document the GNOME CoC lists as inspiration.

---
For a third time now, a CoC is poised to be implemented here without asking contributors.
Certainly my efforts here are lesser than others, but I am meant to review it how?
Given the ability, I would not put up a document virtue-signalling ideas about inclusion when it is all removed from both ability and concent…
Either give everyone the voice they are claimed to agree with, or drive authoritarianism without the rosy words.
I have no problem about others ultimately being in charge, just be honest about it, and what it is.

What isn't a positive experience is having immutable characteristics listed as paramount to codifying conduct.
"Translators" further don't get a say in the mandated "inclusive language", which is _exactly_ what one would expect.
I refuse to leave a review, because I value all contributors, and contribute to libre software for reasons why this could never enter a libre license.

Similar to the haphazardly written and implemented CoC onto the existing project here before, the GNOME one also comes in endless versions, only it doesn't have versioning. The proposed text here isn't the most recent one at GNOME.
Consider the impetus of just updating a document that speaks on behalf of all people that created the platform on which it is done.

https://lwn.net/Articles/366559/
> While there was talk of voting on whether GNOME should leave the GNU project

https://www.yewtu.be/watch?v=a02fdZZOHlQ

Having over a decade of needing to change _a_ CoC means its earlier permutations were all not sufficient, a message overshadowed by an even longer time without one.
I can only hope to avoid replicating the health of the social GNOME ecosystem, and am happy to report this is the case in all circles I frequent.

In advance — hugs to all. I'll take the ban.

mirkobrombin:
I'm not here to discuss the GNOME CoC or to misinterpret it or read between the lines. For information, criticisms or doubts, I invite you to contact the foundation through the dedicated channels.

The purpose of asking for the revision is more a view and a courtesy. I believe we can proceed safely and I again invite you to discuss these things in the dedicated channels and possibly avoiding assumptions.

pietrodicaprio:
I would like to add a personal opinion on the wall text about "how it _can_ be interpreted": a more detailed CoC express better what is accepted and what isn't. Your idea of "almost empty CoC" allow everyone to make his/her own CoC in his/her mind and say "this is ok in my opinion".
The CoC is made to help the community to not "cross the line" and more clear is the line more safer is the place.

jannuary:
okay

PR #1509, opened 2022-05-23
Replace deprecated <mimetypes> tag in metainfo

JakobDev:
See the [specification](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-mimetypes) for reference


mirkobrombin:
Thanks for your contribution

PR #1426, opened 2022-05-06
[WIP] Port to GTK4

jannuary:
Let's get this going. Still very broken and the commits are a mess. Fixes #610, would also port things to new widgetry where possible/or needed.

Checklistt:

- [ ] **python**
  - [x] **misc**
    - [x] remove all gtk_dialog.run's
    - [x] fix InstallerManager.__ask_for_local_resources()
    - [x] running operations are hidden in the main view, no feedback is given to the user

  - [x] **dialogs**
    - [x] crash
    - [x] dlloverrides
    - [x] drives
    - [x] duplicate
    - [x] envvars
    - [x] gamescope
    - [x] generic
    - [x] generic_cli
    - [x] journal
    - [x] onboard
    - [x] rename
    - [x] runargs

  - [x] **views**
    - [x] bottle_dependencies
    - [x] bottle_details
    - [x] bottle_installers
    - [x] bottle_preferences
    - [x] bottle_programs
    - [x] bottle_taskmanager
    - [x] bottle_versioning
    - [x] details
    - [x] importer
    - [x] library
    - [x] list
    - [x] loading
    - [x] new
    - [x] preferences

  - [x] **widgets**
    - [x] component
    - [x] dependency
    - [x] executable
    - [x] importer
    - [x] installer
    - [x] library
    - [x] page
    - [x] program
    - [x] state

  - [x] cli.in (need test)
  - [x] main
  - [x] operation
  - [x] operation_cli
  - [x] window

- [x] **ui**
  - [x] about
  - [x] component-entry
  - [x] dependency-entry
  - [x] details
  - [x] details-bottle
  - [x] details-dependencies
  - [x] details-installers
  - [x] details-preferences
  - [x] details-programs
  - [x] details-versioning
  - [x] dialog-crash-report
  - [x] dialog-dll-overrides
  - [x] dialog-drives
  - [x] dialog-duplicate
  - [x] dialog-env-vars
  - [x] dialog-gamescope
  - [x] dialog-journal
  - [x] dialog-launch-options
  - [x] dialog-rename
  - [x] dialog-run-args
  - [x] dll-override-entry
  - [x] drive-entry
  - [x] env-var-entry
  - [x] importer
  - [x] importer-entry
  - [x] installer-entry
  - [x] library
  - [x] library-entry
  - [x] list
  - [x] list-entry
  - [x] loading
  - [x] new
  - [x] onboard
    - [x] replace bottles-welcome.png draft
  - [x] preferences
  - [x] program-entry
[...]

PR #1565, opened 2022-06-01
Updating the main menu

Knebergish:
It's time to update the main menu items.
In this version, it is proposed to follow the instructions of [Gnome HIG](https://developer.gnome.org/hig/patterns/controls/menus.html#primary-menus) and move the "Preferences" item under the separator. The items "Support us" and "Forums" have been removed, "Documentation" has been renamed to "Help".

On the contrary, I would like to move the "Preferences" item to the very top (because I think this is the most frequently used button in this menu), but who am I to contradict Gnome -_-.


mirkobrombin:
We are making an huge rewrite with #1426

Knebergish:
Good news, thanks!

PR #1622, opened 2022-06-14
fix: appimage test builds

cat-master21:
# Description
First, it Appimage builds failing on Linux Deploy on CI, it also changes the CI to use the latest Ubuntu version instead
of Bionic which does not have gtk4. It also uploads the appimage for example, say, you wanted to test something on a
different branch.

## Type of change
- [x] This change requires a documentation update (alert users that appimages are available)

# How Has This Been Tested?
Log of a run: https://github.com/cat-master21/Bottles/runs/6891750436?check_suite_focus=true
A AppImage available at: https://github.com/cat-master21/Bottles/actions/runs/2499291902 :rocket:!!!

### DIFF
--- a/.github/workflows/test_appimage_build.yml
+++ b/.github/workflows/test_appimage_build.yml
@@ -13,7 +13,7 @@ jobs:
 
       - name: Install Packages
         run: |
-          sudo apt-get install python3-pip python-dev build-essential gettext git libgtk-4-dev -y
+          sudo apt-get install python3-pip python-dev-is-python3 build-essential gettext git libgtk-4-dev -y
           sudo pip3 install --upgrade pip
           sudo pip3 install meson
           sudo pip3 install ninja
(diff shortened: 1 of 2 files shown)


cat-master21:
Here is a picture of boot:
Screenshot from 2022-06-14 19-03-48

mirkobrombin:
Hi thanks for your contribution. I'm sorry but we already have a new GitHub workflow using Appimage builder: https://github.com/bottlesdevs/Bottles/blob/master/.github/workflows/build_pkgs_unstable.yml

We are [working](https://github.com/bottlesdevs/Bottles/issues/1272) to make it works and it's almost done. The [appimage-build.sh](https://github.com/bottlesdevs/Bottles/blob/master/appimage-build.sh) script is a old-bad-practice we used to build the Appimage and it's still there only for development purposes.

PS. libadwaita is a new core dependency since 2022.6.14, so it should be installed too.

Again, I'm sorry to have to "reject" your work. Thanks again <3

PR #1635, opened 2022-06-15
fix: fix appimage for unstable github builds

### LINKED ISSUE #1272: Bottles AppImage

cat-master21:
# Description
This intends to fix appimages for unstable github workflow. In addition I'm also trying to fix the changed dependencies included with the appimage.

Fixes #1272 

## Type of change

# How Has This Been Tested?
Example run: https://github.com/cat-master21/Bottles/actions/runs/2503841056
the appimage uploaded however is currently broken.

### DIFF
--- a/.github/workflows/build_pkgs_unstable.yml
+++ b/.github/workflows/build_pkgs_unstable.yml
@@ -12,23 +12,23 @@ jobs:
       # Dependencies
       # --------------------------------------
       - name: Install build dependencies
-        run: | 
-          apt update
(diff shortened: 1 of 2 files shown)


mirkobrombin:
Bottles 2022.6.14 uses libadwaita 1.2.0 which is not released yet, so we should build it with the Appimage

mirkobrombin:
Nice job 👍

PR #1638, opened 2022-06-15
details: fix title buttons position

nahuelwexd:
# Description
In the details screen, the title buttons were **always** displayed in the content pane. This is not a problem with GNOME's default decoration layout (`appmenu:close`), but in DEs like elementary or KDE, this brings an appearance that is perhaps not as expected:

imagen  
imagen

This PR allows any controls that need to be displayed in the sidebar panel to do so seamlessly:

imagen  
imagen

## Type of change


jannuary:
This breaks when folded, doesn't it?

nahuelwexd:
> This breaks when folded, doesn't it?

Yes, marking as draft to fix it

PR #1645, opened 2022-06-16
chore: cleanup unused appimage building method

cat-master21:
# Description
AppImages used to be built by a shell script but now Bottles's AppImage is built by AppImage-Builder so the old method is no longer needed. This old method is also very broken.

## Type of change
- [x] Cleanup (Old files/code that can be removed without changing anything)


mirkobrombin:
I keep using this (need to be updated) for development purposes.

PR #1665, opened 2022-06-22
Improve terminal emulator priorities

jtojnar:
# Description
Until [XDG Terminal Intent](https://gitlab.freedesktop.org/xdg/xdg-specs/-/merge_requests/46) specification is widely supported, we can only guess what terminal emulator the user prefers.
Let’s look for third-party programs first, since those had to be installed explicitly by the user, and xterm last, as it tends to be present as a fallback.

## Type of change

# How Has This Been Tested?
- [ ] Test A
- [ ] Test B

### DIFF
--- a/src/backend/utils/terminal.py
+++ b/src/backend/utils/terminal.py
@@ -36,18 +36,22 @@ class TerminalUtils:
     }
 
     terminals = [
+        # Part of Flatpak package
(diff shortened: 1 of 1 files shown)


mirkobrombin:
Makes sense, thanks

PR #1668, opened 2022-06-22
Initial vkBasalt dialog

### LINKED ISSUE #1607: [Request]: vkBasalt configuration per bottle

Currently Bottles only uses one vkBasalt config for every bottle, which is located at `~/.var/app/com.usebottles.Bottles/config/vkBasalt/vkBasalt.conf` if using Flatpak. There's no convenient method to use a specific config per bottle.

I propose to create a GUI editor for vkBasalt. vkBasalt has the option to take configs from the `VKBASALT_CONFIG_FILE` environment variable, which gives us the ability to apply specific configs for each bottle.

### Other solutions?

_No response_

I am writing a CLI for vkBasalt, since vkBasalt doesn't have one. We could use this program as a backend to generate configs: https://gitlab.com/TheEvilSkeleton/vkbasalt-cli.

TheEvilSkeleton:
# Description
This MR creates a dialog to configure vkBasalt. It also lets us create vkBasalt configurations per bottle.

Closes #1607 

**Checklist**:
- [x] save state
- [ ] clean up codebase
- [ ] ~~implement [`ListBox`](https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/ListBox.html#Gtk.ListBox) to adjust layer ordering~~ Depends on https://gitlab.gnome.org/GNOME/libadwaita/-/issues/313.
- [x] ~~implement import button to import vkBasalt configuration files ([`FileChooser`](https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/FileChooser.html#Gtk.FileChooser))~~ Blocker: LUT files
- [ ] ~~sliders to configure layers ([`Scale`](https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/Scale.html#Gtk.Scale) and [`Adjustment`](https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/Adjustment.html#Gtk.Adjustment))~~
- [x] input box where the user can manually input amount
- [x] gray every option when Default button is ticked ([`is_sensitive`](https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/Widget.html#Gtk.Widget.is_sensitive))

**Previews**:
https://cdn.fosstodon.org/media_attachments/files/108/516/750/751/087/299/original/995b8d977442c50c.mp4
https://cdn.discordapp.com/attachments/957656402216771604/989175528718209074/2022-06-22_10-29-45.mp4

## Type of change

# How Has This Been Tested?
- [ ] Test A
- [ ] Test B


src/backend/utils/vkbasalt.py
```diff
@@ -0,0 +1,181 @@
...
+                exit(1)
+
+        # --lut-file-path
+        if args.lut_file_path:
+            file.append(f"lutFile = {args.lut_file_path}\n")
+
+        # Output file
+        if args.output:
+            if path.isdir(args.output):
+                f = open(f"{args.output}/vkBasalt.conf", "w")
```
A6GibKm, line 151: Please always use `with open(..., ...) as f:` when doing IO, it takes into account a few things that might fail otherwise, like automatically closing the file when it goes out of scope or when there is a IO error (which is more important).

src/backend/utils/vkbasalt.py
```diff
@@ -0,0 +1,181 @@
...
+            else:
+                stderr.write("Error: No such directory\n")
+                exit(1)
+        else:
+            tmp_dir = "/tmp/vkBasalt.conf"
+            f = open(tmp_dir, "w")
+        output = len(args.effects)
+        if output > 0:
```
A6GibKm, line 159: In python most stuff behave as booleans, you can do `if list:` which should be equivalent to `if len(list) > 0`.

src/backend/utils/vkbasalt.py
```diff
@@ -0,0 +1,181 @@
...
+            else:
+                stderr.write("Error: DLS sharpness must be above 0 and below 1\n")
+                exit(1)
+
+        # --dls-denoise
+        if args.dls_denoise:
+            if 0 <= args.dls_denoise <= 1:
+                file.append(f"dlsDenoise = {args.dls_denoise}\n")
+            else:
+                stderr.write("Error: DLS denoise must be above 0 and below 1\n")
```
A6GibKm, line 81: Warning and error handling in Bottles is done via the logging library, i.e. `logging.error(f"Error: DLS denoise must be above 0 and below 1")`.

src/dialogs/vkbasalt.py
```diff
@@ -0,0 +1,133 @@
...
+    def __update(self, config):
+
+        parameters = config["Parameters"]
+        self.default.set_text(str(parameters["default"]))
+
+    def __idle_save(self, *args):
+
+        config = ManagerUtils.get_bottle_path(self.config)
+
+        class settings:
```
A6GibKm, line 78: Python uses [PEP-8](https://peps.python.org/pep-0008/#class-names) for naming, this should be `Settings`. It might also be a bit cleaner if it is defined outside the scope of `vkBasaltDialog` as its own thing, maybe `class VkBasaltSettings`.

src/ui/dialog-vkbasalt.ui
```diff
@@ -0,0 +1,511 @@
...
+    <property name="modal">True</property>
+    <property name="title" translatable="yes">vkBasalt Settings</property>
+    <child>
+      <object class="GtkBox">
+        <property name="orientation">vertical</property>
+        <child>
+          <object class="AdwHeaderBar">
+            <child type="start">
+              <object class="GtkButton" id="btn_cancel">
+                <property name="label" translatable="yes">Cancel</property>
```
A6GibKm, line 19: It might be a good idea to allow mnemonics here, (you press `alt` and a underline appears bellow a letter of the button so pressing that letter activated the button). 

E.g.

```xml
                <property name="label" translatable="yes">_Cancel</property>
                <property name="use_underline">True</property>
```

src/ui/dialog-vkbasalt.ui
```diff
@@ -0,0 +1,511 @@
...
+                </style>
+              </object>
+            </child>
+          </object>
+        </child>
+        <child>
+          <object class="AdwPreferencesPage">
+            <child>
+              <object class="AdwPreferencesGroup">
+                <property name="description" translatable="yes">Configure vkBasalt settings.</property>
```
A6GibKm, line 36: This should be `title`.

src/backend/utils/vkbasalt.py
```diff
@@ -0,0 +1,181 @@
...
+
+        # Output file
+        if args.output:
+            if path.isdir(args.output):
+                f = open(f"{args.output}/vkBasalt.conf", "w")
+            else:
+                stderr.write("Error: No such directory\n")
+                exit(1)
+        else:
+            tmp_dir = "/tmp/vkBasalt.conf"
```
A6GibKm, line 156: Don't do this, better to create a tmp file either via https://docs.python.org/3/library/tempfile.html or use the user's `~/.cache` dir (in this case it will be under `.var/app/...`), you can load it using `GLib.get_user_cache_dir()`.

src/ui/dialog-vkbasalt.ui
jannuary: is this needed at all? the window title already says "vkbasalt settings", so this text doesn't provide value

jannuary (commented):
Filed some comments regarding some design issues I've noticed.
In addition, as a user, it is not obvious to me what this should even do. I'd like some flavor text explaining different jargon and actual effects of the applied filters.

src/ui/dialog-vkbasalt.ui
```diff
@@ -0,0 +1,511 @@
...
+                            <property name="draw-value">False</property>
+                            <property name="fill-level">1</property>
+                            <property name="has-origin">False</property>
+                            <property name="hexpand">True</property>
+                            <property name="inverted">False</property>
+                            <property name="restrict-to-fill-level">False</property>
+                            <property name="value-pos">bottom</property>
+                            <property name="width-request">280</property>
+                            <property name="adjustment">
+                            <object class="GtkAdjustment" id="cas_sharpness">
```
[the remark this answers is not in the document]
TheEvilSkeleton: Is this good? https://github.com/bottlesdevs/Bottles/pull/1668/commits/bafb8465395988cd90720f0df1b650b976c5170d

src/dialogs/vkbasalt.py
```diff
@@ -75,33 +94,13 @@ def __idle_save(self, *args):
...
-            smaa_max_search_steps = False
-            smaa_max_search_steps_diagonal = False
-            smaa_corner_rounding = False
-            lut_file_path = False
-            exec = False
-
         # Applies default settings and closes dialog.
         if self.default.get_state() is True:
-            settings.default = True
+            VkBasaltSettings.settings.default = True
```
A6GibKm: I think here you want to do
```python
settings = VkBasaltSettings()
settings.settings.default = True
```
python always throws me off with this kind of stuff, I am no sure if using that will alter the default in future dialogs (Open a dialog, set a setting, close the dialog and open another, will the Settings class have the default values or the last setted ones?).

TheEvilSkeleton: I think you got this diff from an older commit. This one works: https://github.com/bottlesdevs/Bottles/pull/1668/commits/f909b38ea63531328a7b1cb60632d02515378ed9.

src/backend/globals.py
```diff
@@ -44,6 +44,16 @@ def get_apps_dir():
         _dir = f"{Path.home()}/.local/share/applications"
     return _dir
 
+def VkBasaltPath():
+    vkbasalt_paths = [
+        "/usr/lib/extensions/vulkan/vkBasalt/share/vulkan/implicit_layer.d/vkBasalt.json",
+        "/usr/share/vulkan/implicit_layer.d/vkBasalt.json"
+    ]
+    for path in range(len(vkbasalt_paths)):
+        if os.path.isfile(vkbasalt_paths[path]):
```
mirkobrombin, line 53: You can simply return this as os.path.file already returns a bool

TheEvilSkeleton: Doing that will only return the first path no? I committed https://github.com/bottlesdevs/Bottles/pull/1668/commits/057d30f01a3b6d83d82a8d4a716924ad569342b1 so it returns False only if those directories are missing.

mirkobrombin: 'os.path.isfile` always return a bool:
```python
>>> import os
>>> print(os.path.isfile("this/path/doesnt/exists"))
False
>>> print(os.path.isfile("/var/home/mirko/this_exists"))
True
>>> 
```

TheEvilSkeleton: That I understand, but what the script is trying to do is check for files in a given list. Your method seems to not work with lists:
```py
>>> import os
>>> install_paths = [ "/usr/lib/extensions/vulkan/vkBasalt/share/vulkan/implicit_layer.d/vkBasalt.json", "/usr/share/vulkan/implicit_layer.d/vkBasalt.json" ]
>>> if os.path.isfile(install_paths):
...     print("Yes")
... 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.9/genericpath.py", line 30, in isfile
    st = os.stat(path)
TypeError: stat: path should be string, bytes, os.PathLike or integer, not list
```

mirkobrombin: You need to keep the loop, just change the return

mirkobrombin: for …:
    return os.path.isfile(..)

instead of
for ..:
    if os.path.isfile(..):
        return True
    return False

TheEvilSkeleton: So something like this?
```py
    vkbasalt_paths = [
        "/usr/lib/extensions/vulkan/vkBasalt/share/vulkan/implicit_layer.d/vkBasalt.json",
        "/usr/share/vulkan/implicit_layer.d/vkBasalt.json"
    ]
    for path in range(len(vkbasalt_paths)):
        return os.path.isfile(vkbasalt_paths[path])
```

TheEvilSkeleton: Won't that only return the very first file?

mirkobrombin: Isn’t already doing it? You are returning inside the loop

TheEvilSkeleton: That was fixed https://github.com/bottlesdevs/Bottles/pull/1668/commits/057d30f01a3b6d83d82a8d4a716924ad569342b1.

TheEvilSkeleton: Correction: I need to fix it :p

TheEvilSkeleton: Correction, I need to fix it :p

TheEvilSkeleton: Fixed. https://github.com/bottlesdevs/Bottles/pull/1668/commits/9111090e0d0f7701050e0f88f0c291f3e8ecabc9

src/backend/utils/vkbasalt.py
TheEvilSkeleton: Bottles won't be using temp files so this is unnecessary. I'm copy pasting the library from https://gitlab.com/TheEvilSkeleton/vkbasalt-cli/-/blob/main/src/vkbasalt.py.

TheEvilSkeleton:
This is how it looks like right now:


https://user-images.githubusercontent.com/50847364/175799425-9f62401d-58e4-4688-be3f-4132584ab2bf.mp4

(Will implement everything suggested at some point.)

TheEvilSkeleton:
> In addition, as a user, it is not obvious to me what this should even do. I'd like some flavor text explaining different jargon and actual effects of the applied filters.

Where should I explain it? I was thinking of adding a Help window like gnome-control-center does. But this is a dialog so I'm not sure how I can explain it. The other option is adding a description in each ActionRow, but the description will be too long.

axtloss:
> > In addition, as a user, it is not obvious to me what this should even do. I'd like some flavor text explaining different jargon and actual effects of the applied filters.
> 
> Where should I explain it? I was thinking of adding a Help window like gnome-control-center does. But this is a dialog so I don't think it'll look good. The other option is adding a description in each ActionRow, but the description will be too long.

Maybe you could add some sort of info button that opens a wiki in the browser which explains everything? Or a description that can be maximized to provide more information.

jannuary:
Will you have any reason to click it though?

jannuary:
Will you have any reason to click it though?

axtloss:
My first instinct when wanting to know more about a specific thing would be to press the button that has an info icon on it. I'd assume most users would do that too.

axtloss:
image

This is what I'd imagine it looks like, when pressed it could open a wiki page or similiar that explains what the value does

src/dialogs/vkbasalt.py
```diff
@@ -228,13 +228,16 @@ def __change_edge_detection_type(self, widget, edge_detection_type):
     def __import_clut(self, *args):
         def set_path(_dialog, response, _file_dialog):
             if response == -3:
-                _file = _file_dialog.get_file()
+                global file_path
```
mirkobrombin, line 240: what's the purpose of making this a global var?

TheEvilSkeleton: It's needed for this: https://github.com/TheEvilSkeleton/Bottles/blob/1f0a808270a7632198a21fac82a5a2a0450363c3/src/dialogs/vkbasalt.py#L192

mirkobrombin: You should avoid using global as much as possible. Global variables make the debug process hard as you lose where/when a value was assigned.

TheEvilSkeleton: What do you suggest instead?

src/dialogs/vkbasalt.py
```diff
@@ -244,24 +240,32 @@ def __import_clut(self, *args):
...
+                img = Image.open(self.lut_file_path)
+
+                def __error_dialog(message):
                     dialog = Gtk.MessageDialog(
                         parent=self.window,
                         transient_for=self,
                         message_type=Gtk.MessageType.ERROR,
                         buttons=Gtk.ButtonsType.CLOSE,
-                        text="Color Lookup Table path must not contain any spaces. Please rename the file to remove all spaces."
+                        text=f"{message}"
```
A6GibKm, line 252: `text=message` directly here.

src/dialogs/vkbasalt.py
```diff
@@ -0,0 +1,265 @@
...
+
+        self.luma.handler_unblock_by_func(self.__change_edge_detection_type)
+        self.color.handler_unblock_by_func(self.__change_edge_detection_type)
+
+    def __import_clut(self, *args):
+        def set_path(_dialog, response, _file_dialog):
+            if response == -3:
+                self.lut_file_path = _file_dialog.get_file().get_path()
+
+                img = Image.open(self.lut_file_path)
```
mirkobrombin, line 244: I may be wrong but this should be achievable using GdkPixbuf too

TheEvilSkeleton: Correct.

A6GibKm: Use Gdk.Texture.new_from_filename¹ or new_from_file. Pixbufs are old gtk3 api. Textures are a subclass of Paintables, GtkImage has a `.set_paintable` method.

¹https://docs.gtk.org/gdk4/ctor.Texture.new_from_filename.html

TheEvilSkeleton: I'm only using it to get the width and height of the image.

A6GibKm: There are methods for paintable to get those, see https://docs.gtk.org/gdk4/vfunc.Paintable.get_intrinsic_height.html

TheEvilSkeleton: Thanks.

TheEvilSkeleton:
> image
> 
> This is what I'd imagine it looks like, when pressed it could open a wiki page or similiar that explains what the option does

Ugh, there's no way to explain these effects in a simple manner. Most of them are undocumented or badly documented.

A6GibKm:
Maybe it's better to just show a popover with some minimal info

TheEvilSkeleton:
See, here's the problem. These effects somewhat solve the same issues, just that they do it differently. I don't think there's a way to provide minimal information, because there will probably be more questions than answers.

Also, how do you suggest implementing the popovers?

A6GibKm:
> See, here's the problem. These effects somewhat solve the same issues, just that they do it differently. I don't think there's a way to provide minimal information, because there will probably be more questions than answers.
> 
> Also, how do you suggest implementing the popovers?

I have this one in Citations, it is a header + brief explanation + code example

```xml
  <object class="GtkPopover" id="popover">
    <child>
      <object class="GtkBox">
        <property name="orientation">vertical</property>
        <property name="spacing">6</property>
        <property name="margin-start">6</property>
        <property name="margin-end">6</property>
        <property name="margin-top">6</property>
        <property name="margin-bottom">6</property>
        <child>
          <object class="GtkLabel">
            <property name="label" translatable="yes">Custom Template</property>
            <property name="xalign">0</property>
            <style>
              <class name="heading"/>
            </style>
          </object>
        </child>
        <child>
          <object class="GtkLabel">
            <property name="xalign">0</property>
            <property name="wrap">True</property>
            <property name="wrap-mode">word-char</property>
            <property name="ellipsize">none</property>
            <property name="lines">4</property>
            <property name="use-markup">True</property>
            <property name="max-width-chars">40</property>
            <property name="label" translatable="yes">Write a custom format for your citations, to insert the contents of a tag &lt;tt&gt;sometag&lt;/tt&gt; write &lt;tt&gt;%SOMETAG%&lt;/tt&gt;. Basic markdown syntax for bold and italics is supported.
            </property>
          </object>
        </child>
        <child>
          <object class="GtkLabel">
            <property name="label" translatable="yes">Example:</property>
            <property name="xalign">0</property>
            <style>
              <class name="heading"/>
            </style>
          </object>
        </child>
        <child>
          <object class="GtkScrolledWindow">
            <property name="height_request">70</property>
            <property name="width_request">300</property>
            <property name="max-content-width">300</property>
            <property name="hscrollbar-policy">never</property>
            <child>
              <object class="GtkTextView" id="help_source_view">
                <property name="hexpand">True</property>
                <property name="wrap_mode">word</property>
                <property name="editable">False</property>
                <property name="monospace">True</property>
              </object>
            </child>
            <style>
              <class name="entry"/>
            </style>
          </object>
        </child>
      </object>
    </child>
  </object>
```

and somewhere else
```xml
                  <object class="GtkMenuButton">
                    <property name="icon-name">dialog-question-symbolic</property>
                    <property name="popover">popover</property>
                    <style>
                      <class name="flat"/>
                    </style>
                  </object>
```

TheEvilSkeleton:
@A6GibKm can you re-review? Apparently the review button isn't working.

src/backend/globals.py
```diff
@@ -43,6 +43,16 @@ def get_apps_dir():
         _dir = f"{Path.home()}/.local/share/applications"
     return _dir
 
+def VkBasaltPath():
+    vkbasalt_paths = [
+            "/usr/lib/extensions/vulkan/vkBasalt/etc/vkBasalt",
+            "/usr/local",
+            "/usr/share/vkBasalt",
+    ]
+    for path in range(len(vkbasalt_paths)):
```
A6GibKm, line 52:
```python
for path in vkbasalt_paths:
    if os.path.exists(path)
    #...
```

src/backend/utils/vkbasalt.py
```diff
@@ -0,0 +1,257 @@
...
+
+def Parse(args):
+    # Apply default settings if possible
+    if args.default:
+        install_paths = [
+            "/usr/lib/extensions/vulkan/vkBasalt/etc/vkBasalt",
+            "/usr/local",
+            "/usr/share/vkBasalt",
+        ]
+        for i in range(len(install_paths)):
```
A6GibKm, line 33: ditto. If you need the index do `for index, item  in enumerate(list)`

src/backend/utils/vkbasalt.py
```diff
@@ -0,0 +1,257 @@
...
+
+def getConfigValue(config, value):
+    with open(config, "r") as f:
+        file = "[config]\n"+f.read()
+        config = configparser.ConfigParser(allow_no_value=True)
+        config.read_string(file)
+        return config['config'].get(value)
+
+def ParseConfig(config):
+    class args:
```
A6GibKm, line 222: `class Args:`

src/dialogs/vkbasalt.py
```diff
@@ -0,0 +1,306 @@
...
+        self.smaa.connect("notify::enable-expansion", self.__check_state)
+        self.clut.connect("notify::enable-expansion", self.__check_state)
+        self.btn_save.connect("clicked", self.__save)
+        self.default.connect("state-set", self.__default)
+        self.luma.connect("toggled", self.__change_edge_detection_type, "luma")
+        self.color.connect("toggled", self.__change_edge_detection_type, "color")
+        self.lut_file_path.connect("clicked", self.__import_clut)
+
+        if os.path.isfile(conf):
+            VkBasaltSettings = ParseConfig(conf)
```
A6GibKm, line 109:
```
vk_basalt_settings = ParseConfig(conf)
```

src/backend/utils/vkbasalt.py
```diff
@@ -0,0 +1,257 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from os import path, environ, system, remove
+from sys import exit
+from shutil import copyfile
+import configparser
+import logging
+
+
+def Parse(args):
```
A6GibKm, line 25: This should be `parse`.

src/dialogs/vkbasalt.py
```diff
@@ -0,0 +1,306 @@
...
+            self.default.set_state(True)
+            self.smaa_edge_detection = "luma"
+            self.cas.set_enable_expansion(False)
+            self.dls.set_enable_expansion(False)
+            self.fxaa.set_enable_expansion(False)
+            self.smaa.set_enable_expansion(False)
+            self.clut.set_enable_expansion(False)
+            self.lut_file_path = False
+
+    def __idle_save(self, *args):
```
A6GibKm, line 168: The callback of a idle_add has to return a boolean, it can be either GLib.SOURCE_CONTINUE or GLib.SOURCE_REMOVE. If you only want to run this once, use the later.

src/dialogs/vkbasalt.py
```diff
@@ -0,0 +1,306 @@
...
+        self.color.handler_unblock_by_func(self.__change_edge_detection_type)
+
+    def __import_clut(self, *args):
+        def set_path(_dialog, response, _file_dialog):
+            if response == -3:
+                self.lut_file_path = _file_dialog.get_file().get_path()
+
+                if self.lut_file_path.split(".")[-1] == "png":
+
+                    texture = Gdk.Texture.new_from_filename(self.lut_file_path)
```
A6GibKm, line 274: This operation can raise a GLib.Error, please use a `try` block.

src/dialogs/vkbasalt.py
```diff
@@ -0,0 +1,306 @@
...
+                        dialog.present()
+
+                    def set_lut_file_path():
+                        if self.input_entry.get_text():
+                            self.lut_file_path = self.input_entry.get_text()
+                        else:
+                            self.lut_file_path = False
+
+                    if " " in self.lut_file_path:
+                        error_dialog("Spaces in File Name", "Color Lookup Table path must not contain any spaces. Please rename the file to remove all spaces.")
```
A6GibKm, line 291: These should be translatable strings. Use `_("Some translatable string")`. Since you added new translatable strings you have to add the file name to `po/POTFILES` (also the UI file).

src/dialogs/vkbasalt.py
```diff
@@ -0,0 +1,306 @@
...
+                        set_lut_file_path()
+                    elif width != height:
+                        error_dialog("Inequal Image Dimension", "Image must have the same dimension.")
```
A6GibKm, line 294: This is hard to read, maybe something more like "Invalid Image Dimension" "The height and width of the image have to coincide".

src/ui/dialog-vkbasalt.ui
```diff
@@ -0,0 +1,819 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!-- Created with Cambalache 0.10.2 -->
+<interface>
+  <!-- interface-name dialog-vkbasalt.ui -->
+  <requires lib="gtk" version="4.0"/>
+  <requires lib="libadwaita" version="1.0"/>
+  <template class="VkBasaltDialog" parent="AdwWindow">
+    <property name="default-width">550</property>
```
A6GibKm, line 8: It might be a good idea to add a default-height.

src/ui/dialog-vkbasalt.ui
```diff
@@ -0,0 +1,819 @@
...
+    <property name="modal">True</property>
+    <property name="title" translatable="yes">vkBasalt Settings</property>
```
A6GibKm, line 10: Add a comment for translators `... name="title" translatable="yes" comments="vkBasalt is a utility name">...` or something of the sort so they don't try to look it up in a dictionary.

src/ui/dialog-vkbasalt.ui
```diff
@@ -0,0 +1,819 @@
...
+                <property name="spacing">6</property>
+                <child>
+                  <object class="GtkButton" id="btn_cancel">
+                    <property name="label" translatable="yes">_Cancel</property>
+                    <property name="use_underline">True</property>
+                    <property name="action_name">window.close</property>
+                  </object>
+                </child>
+                <child>
+                  <object class="GtkShortcutController">
```
A6GibKm, line 28: This has to be a direct child of the AdwWindow, otherwise it wont close the window if the focus is not in the GtkBox which is inside the headerbar.

src/ui/dialog-vkbasalt.ui
```diff
@@ -0,0 +1,819 @@
...
+                          </object>
+                        </child>
+                      </object>
+                    </child>
+                  </object>
+                </child>
+                <child>
+                  <object class="AdwExpanderRow" id="dls">
+                    <property name="show_enable_switch">True</property>
+                    <property name="title" translatable="yes">Denoised Luma Sharpening</property>
```
A6GibKm, line 114: add a comment "Luma is not translatable" or something

src/ui/dialog-vkbasalt.ui
```diff
@@ -0,0 +1,819 @@
...
+            <child>
+              <object class="AdwPreferencesGroup">
+                <property name="description" translatable="yes">Effects are applied according to the list order.</property>
+                <property name="title" translatable="yes">Effects</property>
+                <child>
+                  <object class="AdwExpanderRow" id="cas">
+                    <property name="show_enable_switch">True</property>
+                    <property name="title" translatable="yes">Contrast Adaptive Sharpening</property>
+                    <child>
+                      <object class="AdwPreferencesRow">
```
A6GibKm, line 76: AdwActionRow is already a AdwPreferencesRow, so just remove this object and use the action row directly.

src/ui/dialog-vkbasalt.ui
```diff
@@ -0,0 +1,819 @@
...
+            <property name="max-width-chars">40</property>
+            <property name="label" translatable="yes">FXAA subpixel quality decreases aliasing at the subpixel level. Higher values make the frame softer.</property>
+          </object>
+        </child>
+      </object>
+    </child>
+  </object>
+  <object class="GtkPopover" id="popover_fxaa_quality_edge_threshold">
+    <child>
+      <object class="GtkBox">
```
A6GibKm, line 627: I think one does not need a box here, just put the label directly as a child of the popover with the properties of the box on the label.

On the other hand `orientation=vertical` is wrong here and might cause issues. (Never set orientation on a box containing a single widget)

If you really need a container for a single widget you can use `AdwBin`.

src/ui/dialog-vkbasalt.ui
TheEvilSkeleton: What do you suggest?

src/ui/dialog-vkbasalt.ui
```diff
@@ -0,0 +1,820 @@
...
+                    <child>
+                      <object class="AdwPreferencesRow">
+                        <child>
+                          <object class="AdwActionRow">
+                            <property name="title" translatable="yes">Sharpness</property>
+                            <child>
+                              <object class="GtkMenuButton">
+                                <property name='valign'>center</property>
+                                <property name="tooltip-text" translatable="yes">Select</property>
+                                <property name="icon-name">info-symbolic</property>
```
jannuary, line 85: This icon is broken.
image

src/ui/dialog-vkbasalt.ui
```diff
@@ -0,0 +1,820 @@
...
+                </child>
+              </object>
+            </child>
+            <child>
+              <object class="AdwPreferencesGroup">
+                <property name="title" translatable="yes">Miscellaneous</property>
+                <child>
+                  <object class="AdwActionRow" id="row_disable_on_launch">
+                    <property name="activatable-widget">switch_disable_on_launch</property>
+                    <property name="title" translatable="yes">Disable on Launch</property>
```
jannuary, line 514: What does "Disable on Launch" mean?

src/ui/dialog-vkbasalt.ui
```diff
@@ -0,0 +1,820 @@
...
+                        </child>
+                      </object>
+                    </child>
+                  </object>
+                </child>
+              </object>
+            </child>
+            <child>
+              <object class="AdwPreferencesGroup">
+                <property name="title" translatable="yes">Miscellaneous</property>
```
jannuary, line 510: I don't like having a separate group for one item. I'd just remove the group's title, so that there's still some separation but it isn't as jarring.

src/ui/dialog-vkbasalt.ui
```diff
@@ -0,0 +1,820 @@
...
+                                <property name="tooltip-text" translatable="yes">Select</property>
+                                <property name="icon-name">info-symbolic</property>
+                                <property name="popover">popover_clut</property>
+                                <style>
+                                  <class name="flat"/>
+                                </style>
+                              </object>
+                            </child>
+                            <child>
+                              <object class="GtkButton" id="lut_file_path">
```
jannuary, line 491: Please don't reinvent the wheel here. Just copy the [launch options script picker](https://github.com/bottlesdevs/Bottles/blob/master/src/ui/dialog-launch-options.ui#L56) design.

src/ui/dialog-vkbasalt.ui
```diff
@@ -0,0 +1,820 @@
...
+                                <property name='valign'>center</property>
+                                <property name="tooltip-text" translatable="yes">Select</property>
```
jannuary, line 493: Select what?

src/ui/dialog-vkbasalt.ui
```diff
@@ -0,0 +1,820 @@
...
+                            <child>
+                              <object class="GtkEntry" id="input_entry">
+                                <property name="valign">center</property>
+                                <property name="editable">False</property>
+                              </object>
+                            </child>
+                            <child>
+                              <object class="GtkMenuButton">
+                                <property name='valign'>center</property>
+                                <property name="tooltip-text" translatable="yes">Select</property>
```
jannuary, line 482: Wrong tooltip.

src/ui/dialog-vkbasalt.ui
```diff
@@ -0,0 +1,820 @@
...
+                          </object>
+                        </child>
+                      </object>
+                    </child>
+                  </object>
+                </child>
+                <child>
+                  <object class="AdwExpanderRow" id="clut">
+                    <property name="show_enable_switch">True</property>
+                    <property name="title" translatable="yes">Color LookUp Table</property>
```
jannuary, line 467: "LookUp" is not a word.
```suggestion
                    <property name="title" translatable="yes">Color Lookup Table</property>
```

src/ui/dialog-vkbasalt.ui
```diff
@@ -0,0 +1,820 @@
...
+          </object>
+        </child>
+      </object>
+    </child>
+    <child>
+      <object class="GtkBox">
+        <property name="orientation">vertical</property>
+        <child>
+          <object class="AdwHeaderBar">
+            <property name="show-end-title-buttons">False</property>
```
jannuary, line 27:
```suggestion
            <property name="show-start-title-buttons">False</property>
            <property name="show-end-title-buttons">False</property>
```

Also, why do we even use AdwHeaderBar where it's very much unneccessary, like here? But I digress.

jannuary:
I'm not a fan of popover with a description. It would seem the info is short enough to fit in a description of the ActionRows to me.

A6GibKm:
> I'm not a fan of popover with a description. It would seem the info is short enough to fit in a description of the ActionRows to me.

I have to disagree here, `description` is meant for brief labels

> CAS sharpness increases the sharpness of a frame. Higher values make the frame sharper, whereas values lower than 0 make the frame softer than native.

Just the first part "CAS sharpness increases the sharpness of a frame" is on the limit of what the HIG suggests. Of course if one could produce extremely brief and descriptive descriptions it would be best to have them as a description.

src/ui/dialog-vkbasalt.ui
TheEvilSkeleton: Does it work now?

src/ui/dialog-vkbasalt.ui
TheEvilSkeleton: Essentially, it means that the game is launched without the vkBasalt effects applied. The user would then have to enable it with a key. I removed the section altogether because I don't think it's really useful.

src/dialogs/vkbasalt.py
```diff
@@ -117,11 +118,11 @@ def __init__(self, parent_window, config, **kwargs):
                 self.smaa.set_enable_expansion(False)
             # Check if clut is used.
             if vkbasalt_settings.lut_file_path is None:
-                self.clut.set_enable_expansion(False)
                 self.lut_file_path = False
             else:
-                self.input_entry.set_text(vkbasalt_settings.lut_file_path)
+                self.clut.set_subtitle(vkbasalt_settings.lut_file_path)
                 self.lut_file_path = vkbasalt_settings.lut_file_path
+                self.btn_lut_reset.set_visible(True)
```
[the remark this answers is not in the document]
TheEvilSkeleton: Opened https://github.com/bottlesdevs/Bottles/issues/1808.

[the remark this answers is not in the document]
A6GibKm: It makes no difference whatsoever (I just rechecked GTK's source), one could say it is more idiomatic though.

src/ui/dialog-vkbasalt.ui
TheEvilSkeleton: > I think one does not need a box here, just put the label directly as a child of the popover with the properties of the box on the label.

If we remove the box, it won't have any padding:
image

TheEvilSkeleton: ...I'm not sure why I haven't thought of that.

src/backend/utils/vkbasalt.py
TheEvilSkeleton: Why use this over the one I have right now? I don't think I'll use `index`.

src/ui/dialog-vkbasalt.ui
TheEvilSkeleton: What do you mean?

TheEvilSkeleton:
Wrong button

TheEvilSkeleton:
@A6GibKm can you review?

src/backend/wine/winecommand.py
```diff
@@ -251,6 +251,8 @@ def get_env(self, environment: dict = None, return_steam_env: bool = False, retu
 
         # vkBasalt environment variables
         if params["vkbasalt"] and not self.minimal:
+            if os.path.isfile(os.path.join(ManagerUtils.get_bottle_path(config), "vkBasalt.conf")):
```
mirkobrombin, line 254: just a suggestion, wrap `os.path.join(ManagerUtils.get_bottle_path(config), "vkBasalt.conf")` in a variable and use it for better readability

src/dialogs/vkbasalt.py
```diff
@@ -0,0 +1,326 @@
...
+'''
+
+import os
+from gi.repository import Gtk, GLib, Adw, Gdk
+from bottles.backend.utils.vkbasalt import parse, ParseConfig
+from bottles.backend.utils.manager import ManagerUtils
+from bottles.dialogs.filechooser import FileChooser  # pyright: reportMissingImports=false
+import logging
+
+class vkbasalt_settings:
```
mirkobrombin, line 35: classes should be PascalCase

src/dialogs/vkbasalt.py
```diff
@@ -0,0 +1,326 @@
...
+
+        else:
+            vkbasalt_settings.effects = False
+
+        if self.btn_lut_file_path:
+            vkbasalt_settings.lut_file_path = self.btn_lut_file_path
+
+        vkbasalt_settings.output = conf
+
+        parse(vkbasalt_settings)
```
mirkobrombin, line 229: debugger spotted

src/dialogs/vkbasalt.py
```diff
@@ -0,0 +1,326 @@
...
+            self.btn_save.set_sensitive(False)
+
+    def __default(self, widget, state):
+        self.expander_cas.set_sensitive(not state)
+        self.expander_dls.set_sensitive(not state)
+        self.expander_fxaa.set_sensitive(not state)
+        self.expander_smaa.set_sensitive(not state)
+        self.action_clut.set_sensitive(not state)
+        if state is False:
+            if self.expander_cas.get_enable_expansion() is False and self.expander_dls.get_enable_expansion() is False and self.expander_fxaa.get_enable_expansion() is False and self.expander_smaa.get_enable_expansion() is False and self.btn_lut_file_path is False:
```
mirkobrombin, line 254: consider truncating long lines, PEP8 suggest 72 chars (we don't follow it strictly but this line is very long)
https://peps.python.org/pep-0008/#maximum-line-length

e.g. using `\`
```python
if self.expander_cas.get_enable_expansion() is False \
    and self.expander_dls.get_enable_expansion() is False \
    and self.expander_fxaa.get_enable_expansion() is False \
    and self.expander_smaa.get_enable_expansion() is False \
    and self.btn_lut_file_path is False:
```

src/dialogs/vkbasalt.py
```diff
@@ -0,0 +1,326 @@
...
+                        height = texture.get_height()
+
+                        def set_lut_file_path():
+                            if self.action_clut.get_subtitle():
+                                self.btn_lut_file_path = self.action_clut.get_subtitle()
+                            else:
+                                self.btn_lut_file_path = False
+
+                        if " " in self.btn_lut_file_path:
+                            error_dialog(_("Spaces in File Name"), _("Color Lookup Table path must not contain any spaces. Please rename the file to remove all spaces."))
```
mirkobrombin, line 298:
```python
error_dialog(
    _("Spaces in File Name"), 
    _("Color Lookup Table path must not contain any spaces. Please rename the file to remove all spaces.")
)
```

src/dialogs/vkbasalt.py
TheEvilSkeleton: @A6GibKm hey, you suggested to make this change. What are your thoughts on changing it back?

src/backend/utils/vkbasalt.py
```diff
@@ -0,0 +1,257 @@
...
+
+def parse(args, *arguments):
+    # Apply default settings if possible
+    if args.default:
+        install_paths = [
+            "/usr/lib/extensions/vulkan/vkBasalt/etc/vkBasalt",
+            "/usr/local",
+            "/usr/share/vkBasalt",
+        ]
+        for index, item in enumerate(install_paths):
```
A6GibKm, line 33: you are not using the index it seems, just fo `for item in install_paths`

src/backend/utils/vkbasalt.py
```diff
@@ -0,0 +1,257 @@
...
+                if args.output:
+                    logging.info(f"Outputting file to {file_path}")
+                    copyfile(file_path, path.join(args.output, "vkBasalt.conf"))
+                if args.exec:
+                    environ["ENABLE_VKBASALT"] = "1"
+                    environ["VKBASALT_CONFIG_FILE"] = file_path
+                    system(f"{args.exec}")
+                return
+        logging.error(f"No such path for vkBasalt exists")
+        exit(1)
```
A6GibKm, line 45: In what context this runs? You shouldn't exit a GTK app this way.

TheEvilSkeleton:
Just going to test something.

src/ui/dialog-vkbasalt.ui
```diff
@@ -0,0 +1,691 @@
...
+    </child>
+    <child>
+      <object class="GtkBox">
+        <property name="orientation">vertical</property>
+        <child>
+          <object class="AdwHeaderBar">
+            <property name="show-start-title-buttons">False</property>
+            <property name="show-end-title-buttons">False</property>
+            <child type="start">
+              <object class="GtkBox">
```
A6GibKm, line 30: you don't need a container here, just set the button as a direct child of the headerbar.

src/dialogs/vkbasalt.py
A6GibKm: I think this was a communication issue, yes PascalCase for all classes, always.

TheEvilSkeleton:
Great :/
image

src/dialogs/vkbasalt.py
```diff
@@ -0,0 +1,340 @@
...
+        elif edge_detection_type == "color":
+            self.toggle_color.set_active(True)
+            self.toggle_luma.set_active(False)
+
+        self.toggle_luma.handler_unblock_by_func(self.__change_edge_detection_type)
+        self.toggle_color.handler_unblock_by_func(self.__change_edge_detection_type)
+
+    def __import_clut(self, *args):
+        def set_path(_dialog, response, _file_dialog):
+            if response == -3:
```
A6GibKm, line 280: BURN THIS WITH FIRE. Find the correct response in https://docs.gtk.org/gtk4/enum.ResponseType.html, e.g. Gtk.ResponseType.Ok,but who knows whats the correct number

src/dialogs/vkbasalt.py
```diff
@@ -0,0 +1,340 @@
...
+                                _("The height and width of the image must be equal.")
+                                )
+                            set_lut_file_path()
+                        else:
+                            self.action_clut.set_subtitle(self.btn_lut_file_path)
+                            self.btn_lut_reset.show()
+
+                        self.__check_state()
+
+                    except GLib.Error:
```
A6GibKm, line 320: I suggest to do `except GLib.Error as err` and then do `logging.error("Could not do XXX: ", err.message)` (or whatever fn you use to debug to the cli). I think that logging.debug does NOT accept f-strings.

src/backend/utils/vkbasalt.py
TheEvilSkeleton: It's not supposed to run.

A6GibKm: but then why it is there?

src/ui/dialog-vkbasalt.ui
TheEvilSkeleton: Like this? https://github.com/bottlesdevs/Bottles/pull/1668/commits/522e3ccf733662c0c95dc6ba2fe128f1a6e876db

src/backend/utils/vkbasalt.py
TheEvilSkeleton: It's a library I wrote to make it universal in areas like the CLI. https://gitlab.com/TheEvilSkeleton/vkbasalt-cli

src/ui/dialog-vkbasalt.ui
A6GibKm: Yes. You can also use `type=start` by the way, but it was never needed in the first place, all children of a headerbar are added to the start of the widget unless `=end` is specified.

src/dialogs/vkbasalt.py
A6GibKm: ugh, I thought I had replied to this already. Yeah all classes are PascalCase, *always*, maybe there was a communication issue somewhere.

src/dialogs/vkbasalt.py
A6GibKm: Or 

```python
if (
    x
    and y
    or z
):
```

which is the PEPXYZ way to doing it and what `black` would do

src/dialogs/vkbasalt.py
TheEvilSkeleton: Is this good? https://github.com/bottlesdevs/Bottles/pull/1668/commits/2a980eb741322dc966982c374e381bbc8cdafb4f

Sadly I can't test it because I can't think of a situation where this error will even happen.

axtloss: I tested this by opening the file chooser and deleting the file before selecting it (the file chooser didn't update with the files removal) and I got this error instead:
```
18:03:38 (ERROR) Color Lookup Table path must not contain any spaces. Please rename the file to remove all spaces. 
Traceback (most recent call last):
  File "/app/share/bottles/bottles/dialogs/vkbasalt.py", line 283, in set_path
    self.btn_lut_file_path = _file_dialog.get_file().get_path()
AttributeError: 'NoneType' object has no attribute 'get_path'
```

TheEvilSkeleton: I wasn't able to reproduce it myself, as the file picker updates automatically. I just committed https://github.com/bottlesdevs/Bottles/pull/1668/commits/965398842b6e49f3cc511368f8fe36f6e0e91383, can you try again?

axtloss: It's the same error except the `Color Lookup Table path must not contain any spaces` message doesn't apper

TheEvilSkeleton: Does the file have any spaces? If so, it's normal. The image name should not contain any spaces, and the size of the image must have the same height and width.

axtloss: No both files I tested do not contain any spaces and they both had the same height and width

TheEvilSkeleton: Seems like it works now. @A6GibKm `except GLib.Error as err` doesn't work. I used `AttributeError` instead.

mirkobrombin (approved):
My previous review reports have been corrected, I think we can merge.

TheEvilSkeleton:
I marked it as review. I think it can be merged.

mirkobrombin:
I think it’s ready to merg

Issue #1437, opened 2022-05-08
[Request]: Add a button to enable mangoapp when using Gamescope


VarLad:
### Tell us the problem or your need

Often times, the executable doesn't start when MangoHUD is used with gamescope. There's also the fact that mangoapp has a lot more information when used with gamescope.

### Describe the solution you'd like

Add a button to use gamescope with mangoapp. 

### Additional context and references

More on mangoapp here https://github.com/flightlessmango/MangoHud/issues/625

mirkobrombin:
Thanks for reporting

VarLad:
@mirkobrombin Any update on this? 
Also, NVIDIA's new stable and Gamescope-compatible drivers just got released 
Having mangoapp there would be nice since it has got stuff like showing FSR level

mirkobrombin:
Not yet, I’ve to investigate how to implement it

jntesteves:
I think mangoapp isn't available in Flatpak yet, at least I couldn't find it. Open an issue on MangoHud on Flathub to provide it, I think that'll be required before Bottles can expose it.

I'll try to fix MangoHud on gamescope, so we have at least that before mangoapp.

VarLad:
@jntesteves Thanks for taking the time
I opened an issue yesterday https://github.com/flathub/org.freedesktop.Platform.VulkanLayer.MangoHud/issues/12

TheEvilSkeleton:
https://github.com/flathub/org.freedesktop.Platform.VulkanLayer.MangoHud/pull/14#event-7575424480 was just merged, which gives us the opportunity to use mangoapp. I just merged https://github.com/bottlesdevs/Bottles/pull/2108. I tested locally and it works. I'm still learning how to use mangoapp since the documentation is quite lacking.

PR #1677, opened 2022-06-25
Fix gamescope with MangoHud

jntesteves:
# Description

gamescope crashes with some Vulkan layers like MangoHud. With this change, MangoHud isn't applied to gamescope, but called within gamescope and applied only to the program executing inside it.

This gives us at least working MangoHud within gamescope until #1437 can be done

## Type of change

# How Has This Been Tested?
- [x] Start bottle with both gamescope and MangoHud enabled. Should not crash.
- [x] Start bottle with only MangoHud to verify no regression in what already worked.


mirkobrombin:
Thanks for your contribution

PR #1693, opened 2022-06-29
i18n: Fix Chinese translation not available (#1692)

SwimmingTiger:
# Description

Add `zh_CN`/`zh_HK`/`zh_SG`/`zh_TW` back to `LINGUAS`, so that the Chinese translation can be used by actual users.

Fixes #1692

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?
Please describe the tests that you ran to verify your changes. 
Provide instructions so we can reproduce.

See https://github.com/bottlesdevs/Bottles/issues/1692 and https://github.com/bottlesdevs/Bottles/pull/1515.

Issue #1692, opened 2022-06-29
[Bug]: zh_CN/zh_HK/zh_SG/zh_TW have been removed from LINGUAS, making the Chinese translation unusable


SwimmingTiger:
### Describe the bug

1. `zh_Hans` and `zh_Hant` are not language codes actually used in Linux desktop.

2. `zh_CN`/`zh_HK`/`zh_SG`/`zh_TW` are the language codes actually used in Linux desktop, where `CN`/`SG` use `Hans`, and `HK`/`TW` use `Hant`.

3. I created symlinks for them and added the actual used language code to `LINGUAS`, after this PR, the Chinese translation can be actually used: https://github.com/bottlesdevs/Bottles/pull/1515

4. But unfortunately, this commit was quickly overwritten by https://github.com/bottlesdevs/Bottles/commit/b91be850ced8f47de8df1506ae41e8e3f109e0b4, `zh_CN`/`zh_HK`/`zh_SG`/`zh_TW` was removed from `LINGUAS`, and the Chinese translation is again unavailable to actual users.

### To Reproduce

1. Switch to any Chinese locale.

2. Run Bottles.

3. Bottles are not loaded with Chinese translation.

4. Run this command:
   ```
   env | grep LANG
   ```
    You may get one of the following codes:
    ````
    LANGUAGE=zh_CN
    LANGUAGE=zh_HK
    LANGUAGE=zh_SG
    LANGUAGE=zh_TW
    ````
    You have no chance of getting the following code:
    ````
    LANGUAGE=zh_Hans
    LANGUAGE=zh_Hant
    ````

5. Run this command:
   ```
   cat /etc/locale.gen | grep zh
   ```
   You will get:
   ```
   #lzh_TW UTF-8  
   #zh_CN.GB18030 GB18030  
   #zh_CN.GBK GBK  
   #zh_CN.UTF-8 UTF-8  
   #zh_CN GB2312  
   #zh_HK.UTF-8 UTF-8  
   #zh_HK BIG5-HKSCS  
   #zh_SG.UTF-8 UTF-8  
   #zh_SG.GBK GBK  
   #zh_SG GB2312  
   #zh_TW.EUC-TW EUC-TW  
   #zh_TW.UTF-8 UTF-8  
   #zh_TW BIG5
   ```
   And `zh_Hans` and `zh_Hant` are not among them.

[...]

PR #1712, opened 2022-07-03
Revert "i18n: Fix Chinese translation not available (#1692)"

mirkobrombin:
Reverts bottlesdevs/Bottles#1693


mirkobrombin:
I need to remove those to fix weblate otherwise no translations will works. I'm checking for a solution as weblate seems is not supporting those locales.

PR #1872, opened 2022-07-30
Remove mention of AUR and unofficial packages

TheEvilSkeleton:
# Description
Since we let go of the AUR package, let's stop mentioning it as an official package.

Proposal: I think we should stop mentioning unofficial packages altogether. Flatpak is officially maintained and always in the latest version. Users typically get a subpar experience with distro packages because they have to deal with outdated libs, overlooked dependencies, etc. and may stop using the app and come up with the wrong conclusion.

All in favor? @mirkobrombin @jannuary @axtloss

## Type of change


mirkobrombin:
I agree

axtloss:
i also agree

PR #1889, opened 2022-08-02
Add file preloading/caching 

### LINKED ISSUE #1329: [Request]: add vmtouch, ramdisk or other pre-caching support

With some older games the engines were badly optimized so there is a loading screen after every house, map or screen. I not sure how feasible it is with Bottles and Wine (it looks like vmtouch works with wine) but it would be nice to add some kind of feature that pre-loads the game into memory to cut down on long loading.

if I understand correctly then PlayOnLinux has vmtouch and I seen people talking about it:
https://askubuntu.com/questions/1104882/preload-for-playonlinux-application

Add vmtouch or ramdisk (via Wine?) or some other pre-memory or pre-caching loader where user can specify the game files to load into memory when game launches.

### Other solutions?

N/a

Of course this feature would not work for everyone because you need a lot of unused memory. My computer has 16gb and I don’t think it’s ever going above 12gb used so I might be able to take advantage. Those with 32gb definitely can take advantage.

axtloss:
# Description
This allows bottles to preload files into cache before starting the application, causing faster loading times in games and applications.

Fixes #1329 

## Type of change

# How Has This Been Tested?
- [ ] Test A
- [ ] Test B

---

Currently I only have the actual precaching implemented, freeing the cache after the application quits isn't yet implemented and the ui probably needs some tweaks too.

### DIFF
--- a/com.usebottles.bottles.dev.yml
+++ b/com.usebottles.bottles.dev.yml
@@ -194,6 +194,7 @@ modules:
     sources:
       - type: git
         url: https://github.com/hoytech/vmtouch
+        commit: 8f6898e3c027f445962e223ca7a7b33d40395fc6
     build-commands:
       - make
       - make install PREFIX=/app
--- a/com.usebottles.bottles.dev.yml
+++ b/com.usebottles.bottles.dev.yml
@@ -189,6 +189,14 @@ modules:
 
   # Tools / Codecs
   # ----------------------------------------------------------------------------
+  - name: vmtouch
+    buildsystem: simple
+    sources:
+      - type: git
+        url: https://github.com/hoytech/vmtouch
+    build-commands:
+      - make
+      - make install PREFIX=/app
   - name: vulkan-tools
     buildsystem: cmake-ninja
     config-opts:
--- a/src/backend/wine/winecommand.py
+++ b/src/backend/wine/winecommand.py
@@ -11,7 +11,7 @@ from bottles.backend.utils.manager import ManagerUtils
 from bottles.backend.utils.display import DisplayUtils
 from bottles.backend.utils.gpu import GPUUtils
 from bottles.backend.globals import Paths, gamemode_available, gamescope_available, mangohud_available, \
-    obs_vkc_available
+    obs_vkc_available, vmtouch_available
 from bottles.backend.logger import Logger
 from bottles.utils.threading import RunAsync
 
@@ -508,7 +508,7 @@ class WineCommand:
         return " ".join(gamescope_cmd)
 
     def vmtouch_preload(self):
-        vmtouch_command = "/app/bin/vmtouch"
+        vmtouch_command = vmtouch_available
         vmtouch_flags = "-t -v -l -d"
         vmtouch_file_size = " -m 1024M"
         if self.command.find("C:\\") > 0:
@@ -529,7 +529,7 @@ class WineCommand:
         )
         vmtouch_command = "/app/bin/vmtouch"
         vmtouch_flags = "-e -v"
-        command = vmtouch_command+" "+vmtouch_flags+" "+self.vmtouch_files
+        command = vmtouch_available+" "+vmtouch_flags+" "+self.vmtouch_files
         subprocess.Popen(
             command,
             shell=True,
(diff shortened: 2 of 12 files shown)


src/ui/details-preferences.ui
```diff
@@ -572,6 +572,28 @@
                         </child>
                     </object>
                 </child>
+                <child>
+                  <object class="AdwActionRow">
+                    <property name="activatable-widget">switch_vmtouch</property>
+                    <property name="title" translatable="yes">Vmtouch</property>
+                    <property name="subtitle" translatable="yes">Preload game files into ram</property>
```
jannuary, line 579:
```suggestion
                    <property name="subtitle" translatable="yes">Preload game files into RAM</property>
```

axtloss:
How would I add an option in the settings that allows setting a file size which always is in megabytes (M)?
Currently I just have an AdwEntry widget:
image

but I doubt that it's the right way to do it.

mirkobrombin:
Can this setting be omitted? Maybe using a default?

axtloss:
The default in vmtouch is 500M, but some users may want to cache bigger files, which is why that option would be needed.
Although I could hardcode a higher file size

mirkobrombin:
I think we can set 1000 as standard and wait for user feedbacks without adding more options

axtloss:
I removed the option, the pr should now be read to merge.

jannuary:
I don't like how this expects users beforehand to 
1. Know whether or not they need to enable preloading
2. Know what files need to be preloaded

axtloss:
It would be possible to just enable it by default and have it only preload the main executable (what wine/bottles executes), this would probably avoid taking up too much ram that could cause any issues.
Users with more ram could also enable to preload the whole directory where the application is, which would be more effective, but also take up more ram.

src/backend/wine/winecommand.py
```diff
@@ -506,10 +507,43 @@ def __get_gamescope_cmd(self, return_steam_cmd: bool = False) -> str:
 
         return " ".join(gamescope_cmd)
 
+    def vmtouch_preload(self):
+        vmtouch_command = "/app/bin/vmtouch"
```
mirkobrombin, line 511: I think shutil.which("vmtouch") better fit the case as we can also check it before executing. Also moving to globals.py like other checks, e.g.:
https://github.com/bottlesdevs/Bottles/blob/3ef4024fa2213abba405485aa5ac07fcff4c63b1/src/backend/globals.py#L95
and then check:
https://github.com/bottlesdevs/Bottles/blob/3ef4024fa2213abba405485aa5ac07fcff4c63b1/src/backend/wine/winecommand.py#L430

src/views/bottle_preferences.py
```diff
@@ -55,6 +56,7 @@ class PreferencesView(Adw.PreferencesPage):
     btn_manage_gamescope = Gtk.Template.Child()
     btn_manage_sandbox = Gtk.Template.Child()
     btn_manage_versioning_patterns = Gtk.Template.Child()
+    btn_manage_vmtouch = Gtk.Template.Child()
```
mirkobrombin, line 59: You may want to make the switcher non sensitive if vmtouch is not present, just by using the check mentioned in the previous comment:

https://github.com/bottlesdevs/Bottles/blob/3ef4024fa2213abba405485aa5ac07fcff4c63b1/src/views/bottle_preferences.py#L213

mirkobrombin (commented):
Just some improvements

src/views/bottle_preferences.py
axtloss: But vmtouch would be installed in the flatpak since I added it to the manifest

axtloss: i just remembered that bottles also has packages other than the flatpak, I'll change the code according to your suggestion

com.usebottles.bottles.dev.yml
```diff
@@ -189,6 +189,14 @@ modules:
...
   # Tools / Codecs
   # ----------------------------------------------------------------------------
+  - name: vmtouch
+    buildsystem: simple
+    sources:
+      - type: git
+        url: https://github.com/hoytech/vmtouch
+    build-commands:
+      - make
+      - make install PREFIX=/app
```
TheEvilSkeleton, line 199: Don't forget to add this in the json manifest.

TheEvilSkeleton (changes requested):

TheEvilSkeleton (approved):

TheEvilSkeleton:
Marking it as draft to make it clearer for everyone.

axtloss:
This isn't meant to be a draft anymore, i just forgot to remove the `[WIP]`

PR #1909, opened 2022-08-07
Bugfix glob file copy_dll

siroccal:
The old code tried to copy the dll's over themselves (giving a `SameFileError`) because `_path` and `_dest` would be the same.

Fixes #(issue)
https://github.com/bottlesdevs/dependencies/issues/120

## Type of change

# How Has This Been Tested?
Installing dependencies `d3dcompiler_43`, `d3dcompiler_47`, `d3dx9`, and `d3dx11` now works properly.


mirkobrombin:
Seems you are right but I’ve to figure out why DLLs are being copied in my installation (comparing hash), before merge

Issue #1834, opened 2022-07-24
[Request]: add `MESA_LOADER_DRIVER_OVERRIDE=zink` when launching a game in Gamescope with an Intel iGPU


allaeddineomc:
### Tell us the problem or your need

gamescope is rendering broken graphics on intel IGPUs and adding the variable INTEL_DEBUG=norbc when running gamescope on intel fixes the rendering 

### Describe the solution you'd like

this is an example of how the command should look to make it render properly 
``` INTEL_DEBUG=norbc gamescope '/var/home/edge/Games/steam/Besiege/start.sh' 
```

allaeddineomc:
another solution can be adding a field to gamescope settings where the user can add arguments

allaeddineomc:
@mirkobrombin  , i'm trying to implement this but the app doesn't seem to be building with gnome-builder nor with the included appimage script , anything i can do ?

mirkobrombin:
Check if documentation for manual building the flatpak https://docs.usebottles.com/getting-started/installation

TheEvilSkeleton:
> INTEL_DEBUG=norbc

This seems like a debug flag. I don't want to resort to debug flags as a "solution" for stuff like this. Do you happen to know it's safe to use? Perhaps even cite from someone who is qualified in this domain?

allaeddineomc:
> > INTEL_DEBUG=norbc
> 
> This seems like a debug flag. I don't want to resort to debug flags as a "solution" for stuff like this. Do you happen to know it's safe to use? Perhaps even cite from someone who is qualified in this domain?

i managed to add the flag using flat seal , it's a bit janky solution but it's much better than the default behavior , mango hud isn't functioning with gamescope turned on but it's a fair trade for me

TheEvilSkeleton:
Based on https://github.com/Plagman/gamescope/issues/356#issuecomment-1195054736, you should use `noccs` instead of `norbc`. Can you use MangoHud with Gamescope when you set it to `noccs`?

allaeddineomc:
sorry for taking too long to reply , i didn't got notified , using ```noccs``` renders broken graphics that look different form the broken graphics when ran with no environment variables , ```mango hud``` isn't working with ```noccs``` just like with ```norbc```

allaeddineomc:
gamescope disabled 
[...]

PR #1935, opened 2022-08-13
Force Zink when using an Intel GPU

TheEvilSkeleton:
# Description
This works around an issue with Intel GPUs where they would struggle with OpenGL when using Gamescope. This forces to use Zink (`MESA_LOADER_DRIVER_OVERRIDE=zink`). Zink is a compatibility layer to run OpenGL through Vulkan.

We'll have to see how it behaves, and whether it's worth it or not.

Fixes https://github.com/bottlesdevs/Bottles/issues/1834.
Fixes https://github.com/bottlesdevs/Bottles/issues/1937.

## Type of change
- [x] (Possibly) Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?
- [x] Locally

### DIFF
--- a/com.usebottles.bottles.dev.json
+++ b/com.usebottles.bottles.dev.json
@@ -428,8 +428,8 @@
         {
           "type": "git",
           "url": "https://gitlab.gnome.org/GNOME/libadwaita.git",
-          "tag": "1.2.alpha",
-          "commit": "9150f4016a04064a00733b78391c8e6aad49bee8"
+          "tag": "1.2.beta",
+          "commit": "7387b453a4c93fbb60d68ec009af6f719fe8bab9"
         }
       ],
       "modules": [
--- a/com.usebottles.bottles.dev.yml
+++ b/com.usebottles.bottles.dev.yml
@@ -303,8 +303,8 @@ modules:
     sources:
     - type: git
       url: https://gitlab.gnome.org/GNOME/libadwaita.git
-      tag: 1.2.alpha
-      commit: 9150f4016a04064a00733b78391c8e6aad49bee8
+      tag: 1.2.beta
+      commit: 7387b453a4c93fbb60d68ec009af6f719fe8bab9
     modules:
     - name: libsass
       buildsystem: meson


TheEvilSkeleton:
@xiota since you're more knowledgeable about this than I am, this MR essentially sets `MESA_LOADER_DRIVER_OVERRIDE=zink` when it detects an Intel GPU. Do you think it's sufficient? I don't have an Intel GPU to test with, but we're quite limited with hardware.

Otherwise, we could add a toggle that only appears when using an Intel GPU.

bottles/backend/wine/winecommand.py
```diff
@@ -14,6 +14,7 @@
     obs_vkc_available, vkbasalt_available, vmtouch_available
 from bottles.backend.logger import Logger
 from bottles.frontend.utils.threading import RunAsync
+from bottles.backend.health import HealthChecker
```
mirkobrombin, line 17: I suggest to use the GPUUtills instance defined here
https://github.com/bottlesdevs/Bottles/blob/046944ed86dc633d67589a96ffbb76e7661385e0/bottles/backend/wine/winecommand.py#L147

This provide the full GPU configuration: https://github.com/bottlesdevs/Bottles/blob/046944ed86dc633d67589a96ffbb76e7661385e0/bottles/backend/utils/gpu.py#L89

This allow checking for bot prime and classic setup, e.g. 
https://github.com/bottlesdevs/Bottles/blob/046944ed86dc633d67589a96ffbb76e7661385e0/bottles/backend/wine/winecommand.py#L324

xiota:
Maybe an option could be provided for the user to select a workaround.  Blindly forcing one isn't necessarily a good solution.  Based on a small sample of games I've tried:

* Zink fixes *some* games, but *not* others.
* `INTEL_DEBUG=noccs` (or `norbc`) fixes *some* games that are *not* fixed by Zink.
* Some games are fixed by either zink or noccs, but one solution may perform better than the other.
* Some games are fixed *not* by either.  A workaround has not yet determined.

mirkobrombin:
I wonder if it is not appropriate to let the user manage these workarounds in the Environment Variables section as they are in fact variable.

xiota:
@mirkobrombin 

I do think users should manage the workarounds themselves.  The only problem with setting environment variables normally is gamescope crashes (on my computer).  So the variable has to be set in a subshell:

    gamescope -- sh -c "MESA_LOADER_DRIVER_OVERRIDE=zink game_exec"

xiota:
> Otherwise, we could add a toggle that only appears when using an Intel GPU.

Maybe a drop down box with a selection of known workarounds?

TheEvilSkeleton:
> > Otherwise, we could add a toggle that only appears when using an Intel GPU.
> 
> Maybe a drop down box with a selection of known workarounds?

Personally I prefer fixing the environment variable issue and ask the user to edit the variable that way, rather than adding more options.

xiota:
> Personally I prefer fixing the environment variable issue and ask the user to edit the variable that way, rather than adding more options.

@TheEvilSkeleton  That's a reasonable approach.  Variables might have to be set in a subshell for some other programs too, like mangohud.  Maybe a checkbox, or something, for the user to select where the variable needs to be set?

TheEvilSkeleton:
I rather have it done automatically. Essentially, if the user edits the variable, Bottles will intelligently insert the variable and key into the subshell.

xiota:
> I rather have it done automatically.

I just wonder if there's a scenario where a variable might have to be set globally, rather than in a subshell.  I haven't yet encountered a case where that's been needed though.

TheEvilSkeleton:
I'll let @mirkobrombin handle this MR. Apparently GitHub closes the MR automatically when the commit is on par with `master`.

PR #1964, opened 2022-08-21
Simplification of "Legacy Tools" section

### LINKED ISSUE #1945: Descriptions update for Legacy Tools
Hello all.
I have a suggestion for updating section "Details & Utilities" -> "Tools" -> "Legacy Tools".
Old view:
LegacyTools

My suggestion is:
1) Change "Legacy Tools" to "Legacy Wine Tools"
2) Completely remove the description of the items in this section, because it is completely clear from their names what they are doing.
2.1) Or, if it is necessary to keep these descriptions, simplify them in something like this:
```
Legacy Wine Tools \/

Explorer
    Browse internal files.
Task Manager
    Manage processes.
Debugger
    Debug Wine processes.
Configuration
    Adjust internal settings.
Uninstaller
    Uninstall programms.
Control panel
    Access the internal Control Panel.
```
It seems to me that this section will look cleaner this way.
What do you think about this? If you agree, I can create a PR.

Knebergish:
"Legacy Tools" renamed to "Legacy Wine Tools".
Item subtitles removed.

Fixes #1945 

## Type of change

## How Has This Been Tested?
No tests

### DIFF
--- a/bottles/frontend/ui/details-bottle.ui
+++ b/bottles/frontend/ui/details-bottle.ui
@@ -270,12 +270,11 @@
                 </child>
                 <child>
                     <object class="AdwExpanderRow">
-                        <property name="title" translatable="yes">Legacy Tools</property>
+                        <property name="title" translatable="yes">Legacy Wine Tools</property>
                         <child>
                             <object class="AdwActionRow" id="row_explorer">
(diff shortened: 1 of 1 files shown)


mirkobrombin:
Thanks for your contribution

PR #1989, opened 2022-08-27
Remember last used view in topmost `AdwViewStack`

marhkb:
# Description
Quoting the commit message
```
It makes sense to start the application in the main view where it was
closed. For example, it is plausible that most users set up their
bottles once and then want to land directly in the library view after
launch.

This behavior is also known from other programs such as web browsers or
GNOME Calendar, and it seems that Bottles once handled it this way in
the past. Therefore the gsettings key `startup-view` has now been
(re)activated and represents the last selected view in the main window.
```

## Type of change

# How Has This Been Tested?
- Manually, by building Bottles and testing the changes

### DIFF
--- a/bottles/frontend/windows/main_window.py
+++ b/bottles/frontend/windows/main_window.py
@@ -177,7 +177,17 @@ class MainWindow(Adw.ApplicationWindow):
             self.page_list.search_bar.set_key_capture_widget(self)
             self.btn_search.bind_property('active', self.page_list.search_bar, 'search-mode-enabled',
                                           GObject.BindingFlags.BIDIRECTIONAL)
-            self.stack_main.set_visible_child_name("page_list")
+
+            if self.stack_main.get_child_by_name(self.settings.get_string("startup-view")) is None:
+                self.stack_main.set_visible_child_name("page_list")
+
+            self.settings.bind(
+                "startup-view",
+                self.stack_main,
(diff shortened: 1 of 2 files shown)


mirkobrombin:
I think it's a good implementation, I'll check it asap

bottles/frontend/windows/main_window.py
```diff
@@ -174,10 +174,22 @@ def set_manager(result, error=None):
...
             self.page_list.search_bar.set_key_capture_widget(self)
             self.btn_search.bind_property('active', self.page_list.search_bar, 'search-mode-enabled',
                                           GObject.BindingFlags.BIDIRECTIONAL)
-            self.stack_main.set_visible_child_name("page_list")
+
+            if child := self.stack_main.get_child_by_name(self.settings.get_string("startup-view")):
+                self.stack_main.set_visible_child(child)
+            else:
+                self.stack_main.set_visible_child_name("page_list")
+
```
marhkb, line 192: This is actually obsolete.

marhkb: I would still leave it in, though. This prevents the application from "hanging" when a user manually enters an invalid gsetting value for whatever reason. I will create the bin binding after that though.

mirkobrombin: I agree

PR #2031, opened 2022-09-06
Turkish README

06ergin06:
# Description
Please include a summary of the change and which issue is fixed (if available). 
Please also include relevant motivation and context.
Turkish README added your project. 
Fixes #(issue)

## Type of change

# How Has This Been Tested?
- [x] Test A
- [ ] Test B

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,8 @@
   <a href="https://forums.usebottles.com">Forums</a> · 
   <a href="https://t.me/usebottles">Telegram group</a> · 
   <a href="https://usebottles.com/funding">Funding</a>
+  · 
+  <a href="https://github.com/bottlesdevs/Bottles/blob/master/README.tr.md">Turkish README</a> 
 </div>
 
 <br/>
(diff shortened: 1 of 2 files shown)


jannuary:
Thank you for the MR. 

However, I don't want to set a precedent of having READMEs in different languages, considering there is no way to display the appropriate language, there's no way to keep different languages in sync and that README on a Github repo isn't exaxtly user-targeted documentation.

06ergin06:
> Thank you for the MR.
> 
> However, I don't want to set a precedent of having READMEs in different languages, considering there is no way to display the appropriate language, there's no way to keep different languages in sync and that README on a Github repo isn't exaxtly user-targeted documentation.

In fact, there are many examples of this but if you don't want to, you may not agree.,
I also added it to the main readme file in the Turkish README link.
you know :smile:

jannuary:
It's not a matter of agreeing or disagreeing - this approach cannot scale for more than a couple languages and, in addition, creates a maintenance burden.

We really do appreciate translations, and it would be welcome to have translated documentation, but I nevertheless feel that this is not the right way.

PR #1930, opened 2022-08-11
frontend: Use generic names to improve user-friendliness

TheEvilSkeleton:
# Description
This MR changes most titles and subtitles of preferences, using generic names. Many projects use names that do not really explain their intention, which makes it difficult to understand. This makes it more user friendly as these new titles and subtitles better convey the intention of said preferences.

Fixes https://github.com/bottlesdevs/Bottles/issues/1837.

~~Blocked by https://github.com/bottlesdevs/Bottles/pull/2041 and https://github.com/bottlesdevs/Bottles/pull/2043~~.

Example:

image

## Type of change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

# How Has This Been Tested?
Please describe the tests that you ran to verify your changes. 
Provide instructions so we can reproduce.
- [x] Locally

PR #2037, opened 2022-09-08
Updates the Components section of the Bottle Preferences

orowith2os:
I was looking for a way to change the runner easily a while ago, and missed this option. This would make it more visible, and with a note, so that the user doesn't accidentally miss it.

### DIFF
--- a/bottles/frontend/ui/details-preferences.blp
+++ b/bottles/frontend/ui/details-preferences.blp
@@ -230,7 +230,7 @@ template DetailsPreferences : .AdwPreferencesPage {
 
     [header-suffix]
     Button btn_manage_components {
-      tooltip-text: _("Manage Components Versions");
+      tooltip-text: _("Manage Component Versions");
       valign: center;
       icon-name: "applications-system-symbolic";
 
@@ -240,7 +240,7 @@ template DetailsPreferences : .AdwPreferencesPage {
     }
 
     .AdwExpanderRow exp_components {
-      title: _("Components");
+      title: _("Runner Components");
 
       .AdwActionRow row_runner {
         activatable-widget: "combo_runner";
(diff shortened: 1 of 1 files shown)


mirkobrombin:
The description looks good but the title may be confusing as this menu contains dxvk also which is not really related to the runner

mirkobrombin:
Runner Components it's fine. I think this MR should cooperate with #1930

TheEvilSkeleton:
Honestly, my main gripe with this section is that it's a dropdown. "Components" is pretty vague. I spent 5 minutes trying to change the runner and overlooked Components a couple of times. I think a separate `AdwPreferencesGroup` would make things much easier to find.

mirkobrombin:
Like how it was before the Adw port

TheEvilSkeleton:
Should I create a Components `AdwPreferenceGroup` and submit an MR?

mirkobrombin:
I think we can experiment with the current proposal and see if users complains or still get confused. It’s ok improving the ux but some times it’s also convenient to let users learn how the software works.

TheEvilSkeleton (changes requested):
Can you rebase this? I'm considering merging it.

orowith2os:
I dunno how to rebase asdkmnajlsk;dkasd but hopefully this works

TheEvilSkeleton:
It's expecting `;`: https://github.com/bottlesdevs/Bottles/actions/runs/3091273478/jobs/5001140849#step:5:129.

TheEvilSkeleton (changes requested):
It's expecting `;`: https://github.com/bottlesdevs/Bottles/actions/runs/3091273478/jobs/5001140849#step:5:129.

orowith2os:
bleh, got it. Should be good to go

Issue #2058, opened 2022-09-13
Move from Arch Linux to Flatpak for the test build


mirkobrombin:
As Bottles is primary distributed as a Flatpak, using Arch Linux for the test build may not provide an accurate feedback.
We already have a manifest (.dev) in the main branch.

PR #2072, opened 2022-09-16
CI: Build Flatpak

TheEvilSkeleton:
# Description
This replaces the Arch CI for Flatpak. Additionally, it also creates .flatpak bundles so we can install and test them.

Closes https://github.com/bottlesdevs/Bottles/issues/2058.

## Type of change

# How Has This Been Tested?
- [x] CI


mirkobrombin:
I'll test it asap. Can you re-add [build_pkgs_unstable.yml](https://github.com/bottlesdevs/Bottles/pull/2072/files#diff-00cdf1ce72f8acfa8dd3e16bd2659ae8c9a455b43176edb6a3f4c51bc0d3650e) which is being used as reference for the appimage?

mirkobrombin:
Mmh the CI is failing https://github.com/bottlesdevs/Bottles/actions/runs/3067599688/jobs/4954052973

TheEvilSkeleton:
I fail to understand why it fails. I tried validating appstream locally and there is no error.

axtloss:
This appdata issue happens for me too, I always have to either take out all the url tags or use the flatpak builder flatpak instead of a system package

mirkobrombin:
I think the problem is in the version of the package

axtloss:
I got it working by making it remove the url tags from the appdata: https://github.com/axtloss/Bottles/blob/master/.github/workflows/build_flatpak.yml#L21-L23

TheEvilSkeleton:
This removes many important metadata. This is not normal whatsoever.

axtloss:
From what I understand this is just to see if the build succeeds? I don't think removing some url tags from the appdata would be that big of an issue

TheEvilSkeleton:
Marking it as draft because I don't want to remove those URL tags.

TheEvilSkeleton:
Where do I download artifacts?

axtloss:
It should be on the summary page of the action

TheEvilSkeleton:
Alright, it works! Merging.

PR #2128, opened 2022-10-03
Safely access `program` "dict" keys

davidmartos96:
# Description
I've encountered with a bug when adding a Program via the plus icon. If the program is run right after being created, an error is thrown because the dictionary doesn't contain the `arguments` key.
I'm not sure if this is expected or if the dict should contain the key with an empty string.
Restarting Bottles and running the program seems to provide these defaults so it works ok after a restart.

What I did in the PR is to safely access the dictionary values and provide a default when they don't exist, which fixes the issue. 

## Type of change

### DIFF
--- a/bottles/backend/wine/executor.py
+++ b/bottles/backend/wine/executor.py
@@ -118,8 +118,8 @@ class WineExecutor:
         return cls(
             config=config,
             exec_path=program["path"],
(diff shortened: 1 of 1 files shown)


mirkobrombin:
Thanks for your contribution!

PR #2258, opened 2022-11-13
Fix GStreamer system path, and added a switch in bottle

### LINKED ISSUE #2253: [Bug]: GStreamer path in environment variable seems to not work properly
### Describe the bug

It seems that the variable GST_PLUGIN_SYSTEM_PATH is not populated correctly if a runner bundles it, creating a regression.

On version before commit 40586905bef7876c03d9935b3ba1972472e83309, GStreamer is loaded from the system, which in my case makes some intro videos work. But after this commit, some intro videos crashes (like Cult of the Lamb "Devolver digital" intro, which you can reproduce by downloading the demo on Steam).

Not tested, but it seems that previous commit (34358a9f85f5391212292f3fb9166d01a84b51f3) should work according to the code.

This seems to be due to the fact that the code responsible for modifying the GST_PLUGIN_SYSTEM_PATH does not work as intended.

Currently, GST_PLUGIN_SYSTEM_PATH is set to "gstreamer-1.0", which makes no sense as a path is expected. It then breaks GStreamer.

It should be populated with `os.path.join(runner_path, lib)`, and not the `_root` variable. What would be perfect is to load the library path corresponding to the architecture of the Wine prefix (32bits vs 64bits).

### To Reproduce

1. Download Cult of the Lamb (Demo or not), running with wine-ge (latest works)
2. Wait for the intro video
3. You SHOULD have a "Devolver Digital" intro video, but it doesn't 

### Package

From source

### Distribution

Arch Linux

### Troubleshooting Logs

```shell
None
```

### Display Protocol

_No response_

_No response_

### Additional context

Here is the location of the faulty code: https://github.com/bottlesdevs/Bottles/blob/a2f88fcf9c1c4d119a75766963bbeac037c9e329/bottles/backend/wine/winecommand.py#L249

A hacky patch I wrote:

```python
        gst_plugin_path = []
        for lib in gst_libs:
            if os.path.exists(os.path.join(runner_path, lib)):
                gst_plugin_path.append(os.path.join(runner_path, lib))
        gst_plugin_path.reverse()
        env.add("GST_PLUGIN_SYSTEM_PATH", ':'.join(gst_plugin_path))
```

This will create warnings as GStreamer tries to load the 32bits library while the prefix is only 64 bits. As said earlier, the best way would be to load the library corresponding to the architecture of the prefix.

Kinsteen:
# Description
This fixes #2253, and adds a switch in the Bottle settings to use system or embedded GStreamer.

Screenshot_20221113_211249

## Type of change

# How Has This Been Tested?
Tested with and without on Cult of the Lamb (see issue #2253)


cyberphantom52:
i've had gstreamer related issues before in bottles, seems like that's also related to this.

cyberphantom52:
Relevant code in Proton
https://github.com/ValveSoftware/Proton/blob/4221d9ef07cc38209ff93dbbbca9473581a38255/proton#L1099

Kinsteen:
Interesting! I tried to modify the existing code in Bottles to recreate this behavior. It seems that Proton loads the lib64 path and then the lib path.
In my fix, only the lib64 path is loaded in the case of a 64 bits prefix, maybe this could be modified to match Proton's implementation?

When I tried to also load 32 bits libs, GStreamer was not happy about loading 32 bits, so I'm not sure if it's necessary.

bottles/backend/wine/winecommand.py
```diff
@@ -234,19 +232,21 @@ def get_env(self, environment: dict = None, return_steam_env: bool = False, retu
...
             _path = os.path.join(runner_path, lib)
             if os.path.exists(_path):
                 ld.append(_path)
 
-        for lib in gst_libs:
-            _root = os.path.basename(os.path.join(runner_path, lib))
-            if os.path.exists(os.path.join(runner_path, lib)):
-                env.add("GST_PLUGIN_SYSTEM_PATH", _root)
+        # Embedded GStreamer environment variables
+        if params["embedded_gstreamer"] and not return_steam_env:
```
mirkobrombin, line 244: this param doesn't exists in the bottle configuration

Kinsteen: I added it, as it seemed interesting to have

mirkobrombin: Ok but it should be added to the samples as well https://github.com/bottlesdevs/Bottles/blob/main/bottles/backend/models/samples.py

Kinsteen: Oh sorry, will do!

Is it needed at other places?

mirkobrombin: No, this is the only, in this way Bottles will migrate the configurations to the new layout on next start

Kinsteen: Okay thank you.

With the commit you did before, I would assume you wanted it to be default? I'm not sure what would be the best setting for it

Kinsteen: Also, it seems that they are default parameters in https://github.com/bottlesdevs/Bottles/blob/main/bottles/backend/models/bottle.py, should it also be there?

mirkobrombin: > Also, it seems that they are default parameters in https://github.com/bottlesdevs/Bottles/blob/main/bottles/backend/models/bottle.py, should it also be there?

That class is not yet used in Bottles. 



> Okay thank you.
> 
> With the commit you did before, I would assume you wanted it to be default? I'm not sure what would be the best setting for it

It should be a default if doesn't cause any problem

Kinsteen: > That class is not yet used in Bottles.

Okay, I'm not touching it then.

> It should be a default if doesn't cause any problem

I have not tested extensively, but I'm *guessing* it shouldn't. At least it's easily disabled...
I've updated the original post with a screenshot of the setting description.

mirkobrombin (approved):

Kinsteen:
The last commit should add back 32 bits libraries in a 64 bits bottles, so that 32 bits app can load correct GStreamer architecture.

I have NOT tested it as I don't think I have any 32 bits game using GStreamer, but I based the fix on Proton and Lutris implementation.

Kinsteen:
Yes I can understand. I don't understand the motivation of wine-ge to include gstreamer libs, as they still need GStreamer to be installed on the host if I understand correctly how they work.

Kinsteen:
With more evaluation and debugging in mind, I am still not sure how to proceed with this feature.
Forcing the use of the bundled GStreamer can probably cause problems, if the libs that GStreamer plugins were built against are newer on the host system (can this happen in Flatpak?).
I'm using Arch, with no Flatpak, and the bundled GStreamer of the latest wine-ge is complaining about libFLAC missing for example, but it is there, just a newer version.

In the other hand, having the option in the settings can cause confusions amongst user, as most don't really know what GStreamer is.

What are your thoughts?

mirkobrombin:
I think that in this case the setting must remain, perhaps in the form of an environment variable. I don't think there is a way to figure out when to upload them or not.

Kinsteen:
It's a good idea. Maybe it would be nice to update the wiki to add a troubleshooting section for this? Something like "If videos in game aren't playing"

mirkobrombin:
> It's a good idea. Maybe it would be nice to update the wiki to add a troubleshooting section for this? Something like "If videos in game aren't playing"

I agree

Kinsteen:
To be sure, should it be enabled when launching bottles, like `LOCAL_INSTALLERS`, or in a program variable, named something like `BOTTLES_USE_SYSTEM_GSTREAMER`?

mirkobrombin:
It should be in the program arguments

Kinsteen:
image

Here's the bottle-wise setting, it works. I don't think we can add environment variables program-wise?

mirkobrombin:
Yes via arguments, using the %command% placeholder like steam.  Btw env support is global so your recent fix is ok

Kinsteen:
Okay, then I think it's ready to be merged. For the documentation, should I do a pull request in the documentation repo? Also, in which category should the page go in?

mirkobrombin:
I think FAQ is the one which best suits the case

PR #2266, opened 2022-11-14
Fix Latencyflex not disabling correctly and environment variables

Kinsteen:
# Description
Here is a fix for Latency flex.

First, the function to disable it was not called correctly, resulting in the fact that LatencyFlex was never disabling itself after it was enabled once.

Other fix is that the `VK_ICD_FILENAMES` was not set correctly: currently, it would only be assigned to the LatencyFlex JSON, so it would not have the JSON for the GPU, making games crash at launch. This fixes the issue by changing all the `env.add()` of the variable `VK_ICD_FILENAMES` to `env.concat()`. That way, no code is overwriting the variable, and the games load correctly.

## Type of change

### DIFF
--- a/bottles/backend/wine/winecommand.py
+++ b/bottles/backend/wine/winecommand.py
@@ -344,7 +344,7 @@ class WineCommand:
                     gpu_envs = discrete["envs"]
                     for p in gpu_envs:
                         env.add(p, gpu_envs[p])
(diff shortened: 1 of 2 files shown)


mirkobrombin:
Thanks for your contribution

mirkobrombin (approved):

PR #2289, opened 2022-11-16
Fixed setup of dll overrides on bottle start

asbytes:
# Description
This PR makes 2 changes:
1.  If WINEDLLOVERRIDES environment variable defined in bottle's settings, it is added to final list of DLL overrides which is passed to wine.
2.  MSHTML can be enabled now through bottle's settings interface (DLL Overrides or Environment Variables), when it's needed to some apps.

## Type of change

### DIFF
--- a/bottles/backend/wine/winecommand.py
+++ b/bottles/backend/wine/winecommand.py
@@ -185,7 +185,8 @@ class WineCommand:
 
         # Default DLL overrides
         if not return_steam_env:
-            dll_overrides.append("mshtml=d")
+            if all(not s.startswith("mshtml") for s in dll_overrides):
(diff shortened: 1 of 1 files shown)


bottles/backend/wine/winecommand.py
```diff
@@ -67,7 +67,7 @@ def concat(self, key, values, sep=":"):
             values = [values]
         values = sep.join(values)
 
-        if self.has(key):
+        if self.has(key) and self.__env[key]:
```
mirkobrombin, line 70: I think this is redundant

asbytes: Ok, i removed this check.

PR #2357, opened 2022-12-11
General backend improvements

mirkobrombin:
I did a full review of the Bottles backend to avoid the full rewrite. Here some improvements and code clean-up. Will follow a write-up on what needs a rewrite.


francescomasala:
In my opinion, the if statement in line 94 of the backend/repository.py file can be improved (or it can be removed altogether) since there are no yaml files linked to the bottles version in the mirrors

mirkobrombin:
This is a precaution to serve bottles specific repositories when one or more dependencies/components/programs uses actions not available on older Bottles versions.

francescomasala (approved):

bottles/frontend/utils/connection.py
```diff
@@ -48,15 +49,23 @@ def check_connection(self, show_notification=False):
             return False
 
         try:
-            urllib.request.urlopen('https://repo.usebottles.com/components/index.yml', timeout=5)
+            res = subprocess.run(
+                ['curl', '-Is', 'https://repo.usebottles.com'],
```
Kinsteen, line 53: Is curl assumed to be available on every system?

bottles/frontend/utils/connection.py
```diff
@@ -48,15 +49,23 @@ def check_connection(self, show_notification=False):
...
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL
+            )
+
+            if res.returncode != 0:
+                raise Exception("Connection status: online …")
```
Kinsteen, line 59: Shouldn't it be "offline", as we're checking if the command failed?

bottles/backend/cabextract.py
```diff
@@ -111,10 +111,7 @@ def __extract(self) -> bool:
                     f"-q {self.path}"
                 ]
                 command = " ".join(command)
-                subprocess.Popen(
-                    command,
-                    shell=True
-                ).communicate()
+                subprocess.Popen(command,  shell=True).communicate()
```
Kinsteen, line 114:
```suggestion
                subprocess.Popen(command, shell=True).communicate()
```

bottles/backend/health.py
```diff
@@ -269,6 +269,7 @@ def get_ram_data(self):
 
     def get_results(self, plain: bool = False):
         results = {
+            "Oficial Package": "FLATPAK_ID" in os.environ,
```
Kinsteen, line 272:
```suggestion
            "Official Package": "FLATPAK_ID" in os.environ,
```

bottles/frontend/utils/connection.py
mirkobrombin: It's available in the flatpak which is enough

pietrodicaprio (approved):

PR #2361, opened 2022-12-12
backend: create_desktop_entry: Quote passed links, to allow symbols like &

psydvl:
# Description
### backend: create_desktop_entry: Quote passed links, to allow symbols like &

Fix issue with URLs like `handler://arg1=bla&arg2=bla`, which cause next error
```
/bin/sh: line 1: arg2=bla: command not found
```

Also, there is a typo fix

## Type of change


mirkobrombin:
Is this still a draft?

kirides:
Isn't that an invalid uri scheme?
It should look like the following `scheme://host/path`

where `host` usually is an entrypoint for the handler of `scheme`, like in the case of "bottles" its `run`, e.g. `bottles://run?arg1=v1&arg2=v2` or `bottles://run/some?arg1=v1&arg2=v2`

PR #2400, opened 2022-12-19
Async fetch of components, installers and dependencies

Kinsteen:
# Description
This is a draft PR to implement asynchronous fetch of resources needed by Bottles at startup, to alleviate the long startup problems if you have a bad connection.

The code is very ugly as this is just a concept that could definitely be improved.

I would gladly have any feedback on this, like tests from users, people reviewing the changes...

## Benchmark
I used `trickle` to limit the bandwidth of Bottles. Example: `trickle -s -d 10 bottles`
Here, the startup time is from the instant that the command is typed, to the instant when the bottles show up. It also counts the time when the window is not showing up.

| Without download limit | Startup time | Updated resources time (after startup) |
|--------------------------|--------------|----------------------------------------|
| Without the PR           | ~2s           | 0s                                     |
| With the PR              | -1s           | 0s                                     |

| Download limit of 50KB/s | Startup time | Updated resources time (after startup) |
|--------------------------|--------------|----------------------------------------|
| Without the PR           | 3s           | 0s                                     |
| With the PR              | 1.5s           | 1s                                     |

| Download limit of 10KB/s | Startup time | Updated resources time (after startup) |
|--------------------------|--------------|----------------------------------------|
| Without the PR           | 14s           | 0s                                     |
| With the PR              | 5s           | 10s                                     |

Fixes a lot of issues

## Type of change

# How Has This Been Tested?
Locally, very quickly.


bottles/frontend/utils/threading.py
```diff
@@ -41,7 +41,7 @@ def __init__(self, task_func, callback=None, *args, **kwargs):
             faulthandler.enable()
 
         self.source_id = None
-        assert threading.current_thread() is threading.main_thread()
+        #assert threading.current_thread() is threading.main_thread()
```
mirkobrombin, line 44: If you need to comment this, then something wrong is happening. This line raise an exception when threads are colliding

mirkobrombin:
Thanks for your contribution, I left a comment

bottles/frontend/utils/threading.py
Kinsteen: My understanding is that this line forbids to create a thread inside another thread. Is this what's intended? If yes, may I know why it is needed?

Kinsteen: My first thought that this was to avoid threads launching threads in the case that the parent threads ends before the child, killing it.

But I tested, and even if the parent doesn't join and ends early, the child isn't killed, so I don't think this line is useful.

mirkobrombin: > My understanding is that this line forbids to create a thread inside another thread. Is this what's intended? If yes, may I know why it is needed?

Because Gtk is not thread safe. This line prevents you from accessing the main thread from a second one.

Kinsteen: I don't really understand how this line prevents that? It just blocks threads to be created by another thread. Unless GTK is run in it's own thread, and threads created from the GTK thread can access it?
If this is why, I feel like this is not a really good solution as it blocks a lot of simplicity of creating threads from the GTK thread, even if those children don't do anything related to GTK (like here for example).

I don't even know how the GTK thread could then create threads from the main thread without great refactoring of the whole architecture.

Kinsteen:
I wrote a proof-of-concept of a loading screen in the preferences, when the loading is not quick enough.

https://user-images.githubusercontent.com/5410562/208489613-d7abbfc0-0c8f-4acc-a17e-0304bb991a55.mp4

What do you all think?

francescomasala:
I tested with @Kinsteen for any errors from cloudflare worker but luckily everything went well and cloudflare did not return any errors / report suspicious attacks

Kinsteen:
@mirkobrombin I have triggered a race condition with the threading and calls to GTK, resulting in a crash. I understand more precisely the issue you mentioned.

I'm not sure what's the best way to deal with this. I feel like Bottles could really use a lot of asynchronous code to speedup lots of things while still having a clean and responsive UI, but with the `assert` removed, we need to be very careful about how we implement async code, especially with the UI.

Maybe a rewriting of threading.py could be done so that it launches threads only from the main thread?

mirkobrombin:
We cannot obtain a true async approach in Bottles, that's why I developed the RunAsync class, so:
main thread -> RunAsync -> callback -> main thread

I taken a look at your code and the whole thing should be feasible with the RunAsync approach, just it is a bit tricky and frustrating.

francescomasala:
It's my opinion but I think that in this PR it would be better to implement the downloads "section" with runasync and create another PR to implement the rest calmly and with the necessary attention

Kinsteen:
@mirkobrombin What would be perfect would be: random thread -> main thread -> RunAsync -> callback ... That way we gain the ability to launch threads from anywhere, but still *from* the main thread so we don't have problems if we do UI updates. Just to be certain, it's okay to do UI updates from a function run through RunAsync if it's called from the main thread?

@francescomasala Yes that's probably what will happen, but for now the code for the download is dependent on being able to launch threads from other threads, which is not good.

Kinsteen:
I understand a bit more how it's supposed to work. RunAsync should only be run from the main thread, to do expensive computation. If we want to update the UI, we should use GLib.idle_add from within the created thread, right? That way we won't have any race conditions.

It is somewhat annoying, but I think it's possible to write a lot more async code with this, but it might be less readable.

mirkobrombin:
> I understand a bit more how it's supposed to work. RunAsync should only be run from the main thread, to do expensive computation. If we want to update the UI, we should use GLib.idle_add from within the created thread, right? That way we won't have any race conditions.
> 
> It is somewhat annoying, but I think it's possible to write a lot more async code with this, but it might be less readable.

Exactly

Kinsteen:
Is there a reason why we couldn't run idle_add from whatever thread we create? Like main thread -> child -> child -> GLib.idle_add ? I feel like there's no indication that this couldn't work 🤔

Kinsteen:
I would like some feedback. Currently, when opening a Bottle everything loads at once, creating a small freeze. I'm not sure this is the smoothest way to do things. I've implemented a way to make the dependencies page (for example) load when you click on the button, making it load there. As it is still a bit expensive to load everything, a loading page is shown while everything is computed (or fetched if your connection is slow).

What do you all think? Is it better to have a small freeze when opening the bottle, or loading lazily stuff this way is good too?

https://user-images.githubusercontent.com/5410562/209454082-ab26cdcb-4f0c-4d5a-9cbd-195e06a8feec.mp4

TheEvilSkeleton:
This feels SO MUCH better! I have some feedback: when it's loading dependencies, it would be nice to have a loading screen. Once everything is finished loading, you can show everything up.

Kinsteen:
> when it's loading dependencies, it would be nice to have a loading screen. Once everything is finished loading, you can show everything up.

What do you mean by loading screen? With this, there's already a "status page" that says that everything is loading, and then it shows all the dependencies. Do you have something else in mind?

TheEvilSkeleton:
I edited the comment and added a video to demonstrate what I'm saying. Can you refresh the page?

Kinsteen:
Ah yes sorry, it didn't update for me.

Yes I think it would definitely makes it smoother! I don't really know how to do that, but I can take a look at your code there 🙂

TheEvilSkeleton:
No worries, here are the code:
- https://gitlab.com/TheEvilSkeleton/Upscaler/-/blob/main/upscaler/gtk/window.blp
- https://gitlab.com/TheEvilSkeleton/Upscaler/-/blob/main/upscaler/window.py

Kinsteen:
Thanks!

Kinsteen:
Here's the page with the spinner loader:

https://user-images.githubusercontent.com/5410562/209467135-b14c9209-68d4-4f52-b7c1-a9b32c328d95.mp4

Should we write a message under it, or is it clear enough?

TheEvilSkeleton:
> Should we write a message under it, or is it clear enough?

It's clear enough, tbh. I love it!

bottles/backend/managers/manager.py
```diff
@@ -62,6 +62,9 @@
 from bottles.backend.wine.reg import Reg
 from bottles.backend.wine.regkeys import RegKeys
 from bottles.backend.wine.winepath import WinePath
+from bottles.frontend.utils.threading import RunAsync
+
+from bottles.frontend.utils.threading import RunAsync
```
TheEvilSkeleton, line 67: This is already being imported.
```suggestion
```

TheEvilSkeleton (changes requested):

Kinsteen:
> It's clear enough, tbh. I love it!

Okay great, I'll let this like that then!

To be clear, this PR is not to be merged! It's a bit of a trunk right now for all stuff I'm doing with async, but everything will go in another proper PR when #2412 is merged. I'm not sure if I should close it already, as letting it open makes the CI run and create test packages if people want to test it

Kinsteen:
All the proper work will now be in #2444

Issue #2403, opened 2022-12-19
[Bug]: First start stuck on Almost done


janltu:
### Describe the bug

I just downloaded bottles on flatpak started it for the first time and its stuck on almost done.

Console log:

[...]
18:19:34 (INFO) Renaming [dxvk-2.0-1-18b0ef6.tar.gz] to [dxvk-2.0-1-18b0ef6.tar.gz]. 
18:19:34 (ERROR) Downloaded file [dxvk-2.0-1-18b0ef6.tar.gz] looks corrupted. 
18:19:34 (ERROR) Source cksum: [f7121c4c7bebc257fbb1626690a7dd06] downloaded: [81d4f690b6cd535eaf642ab4385f62c2] 
18:19:34 (ERROR) Removing corrupted file [dxvk-2.0-1-18b0ef6.tar.gz]. 
18:19:34 (ERROR) Extraction failed! Archive ends earlier than expected. 
18:19:34 (WARNING) No dxvk found. 

happens everytime already tried restarting

### To Reproduce

1. Click on Continue on first start 
2. Be stuck

### Package

Flatpak from Flathub

### Distribution

pop_OS!

### Troubleshooting Logs

```shell
cant click installation has not finished
```

ijaure299:
same, install Pop OS and I have exaclly the same problem

ZeitDev:
steamOS, same error, same log

Zardeth:
Same issue on SteamOS

EnfrigFalco:
Fresh install across a desktop and laptop both Fedora 37 can confirm stuck on "Almost Done" tried both flatpak and repo, both computers. Bottles is definitely bugged

bunglesss:
I'm getting this as well on latest Fedora, it just goes round and round saying runtime.tar.gz is corrupted. Fresh install of Fedora from Live CD with nothing else installed on it but Bottles.

Kinsteen:
Can you all send logs about which file exactly was corrupted please?

bunglesss:
Hi mate, I just tried to get the log off that box, but not winning. Its basically downloading runtime.tar.gz over and over again saying its corrupted and the archive ends too soon, then eventually Bottles blows up and crashes.

bunglesss:
[temp.log.zip](https://github.com/bottlesdevs/Bottles/files/10270830/temp.log.zip)

bunglesss:
Oh, was installed from RPM not Flatpak here by the way.

[...]

PR #2407, opened 2022-12-20
fix: ComponentManager.install() deadlock

StoneMoe:
# Description
My internet sucks, manifest fetching and file download randomly failed.
when `ComponentManager.install()` failed with downloading, it won't pop the failed installation from `self.__queue`, next call to `ComponentManager.install()` will keep waiting for empty queue forever.
then Onboard Dialog will stuck at "Almost Done" forever.

## Type of change

# How Has This Been Tested?
- [x] Tested locally

### DIFF
--- a/bottles/backend/managers/manager.py
+++ b/bottles/backend/managers/manager.py
@@ -370,7 +370,7 @@ class Manager:
         tmp_runners = [x for x in self.runners_available if not x.startswith('sys-')]
 
         if len(tmp_runners) == 0 and install_latest:
-            logging.warning("No managed runners found.")
+            logging.warning("No runners found.")
 
             if self.utils_conn.check_connection():
                 # if connected, install the latest runner from repository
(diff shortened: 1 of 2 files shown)


StoneMoe:
Screenshoot by moobell#1174 on Discord

image

StoneMoe:
This should able to fix https://github.com/bottlesdevs/Bottles/issues/2403

bottles/backend/managers/manager.py
```diff
@@ -370,7 +370,7 @@ def check_runners(self, install_latest: bool = True) -> bool:
         tmp_runners = [x for x in self.runners_available if not x.startswith('sys-')]
 
         if len(tmp_runners) == 0 and install_latest:
-            logging.warning("No runners found.")
+            logging.warning("No managed runners found.")
```
jannuary, line 373: Doesn't seem relevant?

jannuary (changes requested):

bottles/backend/managers/manager.py
StoneMoe: > Doesn't seem relevant?

This log confused me becuase it found sys runner but still says "No runner found", so I made it more clear.
I can revert it if you like.

jannuary: It would be better in a separate PR.

Kinsteen:
Looks good, I also stumbled upon problems if Bottles fails to download components.

jannuary (approved):

PR #2430, opened 2022-12-22
misc: Update and improve packaging

TheEvilSkeleton:
# Description
- Remove unnecessary build scripts and manifests from formats we don't intend to use anytime soon.
- Add Devel icons and name
- Update VERSION_UPDATE.md
- Improve shebang in install.sh
- Dynamically get app version, to avoid the need to edit multiple source files

## Type of change
- [x] Misc

# How Has This Been Tested?
- [x] CI and locally


Kinsteen:
I think it would be interesting to log the version of Bottles on start, around https://github.com/bottlesdevs/Bottles/blob/main/bottles/frontend/windows/main_window.py#L116, as if Bottles crashes on start we could know the version without asking if the person having issue provides log.

If the package is built with the CI, will it also have the "devel" tag on the version?

TheEvilSkeleton:
> I think it would be interesting to log the version of Bottles on start, around https://github.com/bottlesdevs/Bottles/blob/main/bottles/frontend/windows/main_window.py#L116, as if Bottles crashes on start we could know the version without asking if the person having issue provides log.

This is out of scope of this MR. It'd be better to discuss about it in an issue or MR.

> If the package is built with the CI, will it also have the "devel" tag on the version?

Yes, that's the goal. The Flatpak manifest of this repo uses the config option `-Ddevel=true`, which (based on my testing) should do everything for you. If you build locally or install from CI artifacts, you should see something like this:

image

mirkobrombin (approved):

PR #2412, opened 2022-12-20
feat: async fetch of components, installers and dependencies

Kinsteen:
# Description
This is the splitting of PR #2400. This part is only the "backend" part, with only the async fetching. No UI has been updated, so for example preferences menu will be empty until the components are fetched AND the menu is closed and reopened.

## Benchmark
I used `trickle` to limit the bandwidth of Bottles. Example: `trickle -s -d 10 bottles`
Here, the startup time is from the instant that the command is typed, to the instant when the bottles show up. It also counts the time when the window is not showing up.

| Without download limit | Startup time | Updated resources time (after startup) |
|--------------------------|--------------|----------------------------------------|
| Without the PR           | ~2s           | 0s                                     |
| With the PR              | -1s           | 0s                                     |

| Download limit of 50KB/s | Startup time | Updated resources time (after startup) |
|--------------------------|--------------|----------------------------------------|
| Without the PR           | 3s           | 0s                                     |
| With the PR              | 1.5s           | 1s                                     |

| Download limit of 10KB/s | Startup time | Updated resources time (after startup) |
|--------------------------|--------------|----------------------------------------|
| Without the PR           | 14s           | 0s                                     |
| With the PR              | 5s           | 10s                                     |

## Type of change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

PR #2444, opened 2022-12-26
frontend: smoother UI with async code

Kinsteen:
# Description
This PR is the continuation of the work in #2400 and #2412, focusing here on the UI side of the changes.

Features:
- Smoother loading of Bottles and pages switching
- Dependencies shows a loading spinner instead of freezing
- Preferences pages also shows loading spinners on runners and DLLs
- Better settings pages when in offline mode

### Demos
https://user-images.githubusercontent.com/5410562/209467135-b14c9209-68d4-4f52-b7c1-a9b32c328d95.mp4

**Simulated slow fetching of components:**

https://user-images.githubusercontent.com/5410562/209553325-88c5ad09-9e00-4831-bdef-01b5abcebca5.mp4

**Preferences when in offline mode**

Screenshot_20221226_165922

## Type of change


mirkobrombin (approved):

bottles/backend/managers/manager.py
```diff
@@ -268,37 +262,52 @@ def check_app_dirs(self):
...
+            self.supported_runtimes = catalog["runtimes"]
+            self.supported_winebridge = catalog["winebridge"]
+            self.supported_dxvk = catalog["dxvk"]
+            self.supported_vkd3d = catalog["vkd3d"]
+            self.supported_nvapi = catalog["nvapi"]
+            self.supported_latencyflex = catalog["latencyflex"]
+            RepoStatus.repo_done_operation("components.organizing")
+        RunAsync(_run)
 
     def organize_dependencies(self):
```
Kinsteen, line 284: Could it be interesting to write a decorator to make this function run asynchronously without having to declare a function and do RunAsync? It could be used a lot in other places I think, and would make the code clearer in my opinion, but it could also make it too abstracted and prone to errors then.

What do you think?

Something like:
```python
@async
def organize_dependencies(self):
    """Organizes dependencies into supported_dependencies."""
    RepoStatus.repo_wait_operation("dependencies.fetching")
    ...

```

mirkobrombin: I tried to make it many times without success

Kinsteen:
I feel like this PR is ready for a merge, but as the project is going towards a bug fixing phase, should this be postponed?

Issue #2491, opened 2023-01-06
[Bug]: wont install d3dx9, d3dcompiler_43, d3dcompiler_47 and maybe other Dependencies


dkeruza-neo:
### Describe the bug

When creating a bottle, the file d3dx9.tar.xz is not extracted inside the prefix, windows/system32/

Wont even won't reinstalling the dependency.

### To Reproduce

Create a bottle or reinstall the dependency.

### Package

Flatpak from GitHub Artifacts

### Distribution

Arch

### Troubleshooting Logs

```shell
Official Package: false
Version: 2022.12.14.1
DE/WM: gnome
Display:
    X.org: true
    X.org (port): :0
    Wayland: true
Graphics:
    vendors:
        amd:
            vendor: amd
            envs:
                DRI_PRIME: '1'
            icd: /usr/share/vulkan/icd.d/radeon_icd.x86_64.json:/usr/share/vulkan/icd.d/radeon_icd.i686.json
    prime:
        integrated: null
        discrete: null
Kernel:
    Type: Linux
    Version: 6.0.12-arch1-g14-2
Disk:
    Total: 1496019107840
    Free: 944889794560
RAM:
    MemTotal: 14.9GiB
    MemAvailable: 8.0GiB
Bottles_envs: null
Tools and Libraries:
    cabextract: true
    p7zip: true
    patool: true
    glibc_min: '2.36'
    icoextract: true
    pefile: true
    orjson: true
    markdown: true
    ImageMagick: true
    FVS: true
    xdpyinfo: true
```

### Additional context

It works from the flatpak version, but there are other issues with it.

Kinsteen:
When you say the version from Github artifacts, which one exactly are you talking about?

dkeruza-neo:
> When you say the version from Github artifacts, which one exactly are you talking about?
> 
> Also, what do you mean it's not copying DLLs? Can you provide logs when installing a dependency?

I pulled the source with git pull, I tried downloading the zip file from github and also built from sources with yay in Arch.
I open the bottle, click on dependencies, I remove and reinstall for example d3dcompiler_43. GUI says its succefully installed, but checking the prefix folder/windows/system32 and the d3dcompiler_43.dll has not been replaced.
How do I get logs?

Kinsteen:
To get logs, simply launch Bottles from the command line.

On my side using latest sources, it works correctly: stubs have a size of 298.6 kbytes, and the real DLLs have a size > 2MB.

[...]

PR #2498, opened 2023-01-07
Addresses bug from patoolib

dkeruza-neo:
# Description
Fixes a bug in "patoolib" that does not extract .tar.xz files correctly if you have xz or 7z installed in your system.
Also added a check to return False when a source DLL is being copied to the wine prefix and the source file does not exist. 

Fixes https://github.com/bottlesdevs/Bottles/issues/2491

## Type of change

# How Has This Been Tested?
To reproduce, install xz or 7z in your distro.

### DIFF
--- a/bottles/backend/managers/dependency.py
+++ b/bottles/backend/managers/dependency.py
@@ -501,10 +501,14 @@ class DependencyManager:
 
             os.makedirs(archive_path)
             try:
-                patoolib.extract_archive(
-                    os.path.join(Paths.temp, file),
-                    outdir=archive_path
-                )
+                ext_path = patoolib.extract_archive(os.path.join(Paths.temp, file), outdir=archive_path)
+                ext_file = ext_path + '/' + os.path.basename(ext_path)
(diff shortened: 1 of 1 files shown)


bottles/backend/managers/dependency.py
```diff
@@ -552,6 +556,9 @@ def __step_copy_dll(self, config: dict, step: dict):
         try:
             if "*" in step.get('file_name'):
                 files = glob(f"{path}/{step.get('file_name')}")
+                if not files:
+                    logging.info(f"File/s not found in {path}")
```
TheEvilSkeleton, line 560:
```suggestion
                    logging.info(f"File(s) not found in {path}")
```

TheEvilSkeleton (changes requested):

TheEvilSkeleton:
I rather not be the one to approve or reject it, as I don't have the time to test right now. I think someone else should review it.

mirkobrombin (approved):

TheEvilSkeleton (approved):

PR #2561, opened 2023-01-17
fix: remove add to library button if already in library

### LINKED ISSUE #2459: details: "Add to library" always available
It's available, even if the program is in the library. Maybe a checkbox would be the easiest solution?

image

Kinsteen:
# Description
Title is quite explanatory

Fixes #2459

## Type of change

# How Has This Been Tested?
- [X] locally


TheEvilSkeleton (changes requested):
The only problem I've encountered is that the "Add to Library" button is still present until you go back and reopen the bottle. I think it should be gone right after it is done adding to the library.

TheEvilSkeleton (approved):

PR #2593, opened 2023-01-20
Don't enable Steam Runtime when switching to `wine-ge-proton`

SuperSamus:
# Description
`wine-ge-proton` contains `proton` in its name, which caused Bottles to enable the Steam Runtime.

## Type of change

### DIFF
--- a/bottles/backend/managers/manager.py
+++ b/bottles/backend/managers/manager.py
@@ -345,17 +345,16 @@ class Manager:
 
         # lock winemenubuilder.exe
         for runner in runners:
-            winemenubuilder_paths = [
-                f"{runner}lib64/wine/x86_64-windows/winemenubuilder.exe",
-                f"{runner}lib/wine/x86_64-windows/winemenubuilder.exe",
-                f"{runner}lib32/wine/i386-windows/winemenubuilder.exe",
-                f"{runner}lib/wine/i386-windows/winemenubuilder.exe",
-            ]
-            for winemenubuilder in winemenubuilder_paths:
-                if winemenubuilder.startswith("Proton"):
-                    continue
-                if os.path.isfile(winemenubuilder):
-                    os.rename(winemenubuilder, f"{winemenubuilder}.lock")
+            if runner not in self.supported_proton_runners:
+                winemenubuilder_paths = [
+                    f"{runner}lib64/wine/x86_64-windows/winemenubuilder.exe",
+                    f"{runner}lib/wine/x86_64-windows/winemenubuilder.exe",
+                    f"{runner}lib32/wine/i386-windows/winemenubuilder.exe",
+                    f"{runner}lib/wine/i386-windows/winemenubuilder.exe",
+                ]
+                for winemenubuilder in winemenubuilder_paths:
+                    if os.path.isfile(winemenubuilder):
+                        os.rename(winemenubuilder, f"{winemenubuilder}.lock")
 
         # check system wine
         if shutil.which("wine") is not None:
--- a/bottles/backend/runner.py
+++ b/bottles/backend/runner.py
@@ -93,7 +93,7 @@ class Runner:
         the host system. There are some exceptions, like the Soda and Wine-GE runners,
         which are built to work without the Steam Runtime.
         """
-        if "proton" in runner.lower() and RuntimeManager.get_runtimes("steam"):
+        if runner in manager.supported_proton_runners and RuntimeManager.get_runtimes("steam"):
             manager.update_config(config, "use_steam_runtime", True, "Parameters")
 
         return Result(


Kinsteen (dismissed):

TheEvilSkeleton:
If I remember right, this was actually intended, because it's made for Steam. @mirkobrombin thoughts?

Kinsteen:
@TheEvilSkeleton No, Proton-GE is intended for Steam. Wine-GE can and should be used without Steam. I don't know why "-proton" was added, but it definitely shouldn't force enable the Steam runtime

TheEvilSkeleton:
Wait, what is this one? I know of wine-ge-custom and proton-ge-custom. I've never heard of wine-ge-proton.

Kinsteen:
Because it doesn't exist, this is Wine-GE aka wine-ge-proton. For some reason, @koplo199 added "proton" to the name of Wine-GE here: https://github.com/bottlesdevs/components/pull/190

Why was this change needed?

koplo199:
@Kinsteen Interesting, I originally changed it because:

1. It is the official name (see [titles](https://github.com/GloriousEggroll/wine-ge-custom/releases): it is `Wine-GE-Proton7-35 Released`, not `Wine-GE-7-35 Released`)
2. It is consistent with other components/runners names: they all fit the `{prefix}-{tag}` naming scheme, including `wine-ge`: `{wine}-{GE-Proton7-35}` (using again the official GitHub repository [tags](https://github.com/GloriousEggroll/wine-ge-custom/tags))
3. Any attempt to use the old name would have required a special handling/tweak/hack just to support it, and the more hack there is, the more difficult it becomes to maintain.

To me, the bug here is on the core side and not on the component side: whether the runner is a proton/wine one shouldn't be guessed by checking a list of hard-coded names when the components [yml files](https://github.com/bottlesdevs/components/blob/main/input_files/10-GE-Proton.yml#L2-L3) already specify `Sub-category: proton` or `Sub-category: wine`.

I see that in [manager.py](https://github.com/bottlesdevs/Bottles/blob/main/bottles/backend/managers/manager.py) there is:
https://github.com/bottlesdevs/Bottles/blob/0bb66aa328e7914a90c248826797d077dd56af90/bottles/backend/managers/manager.py#L100-L101
Could we instead of hard-coding the name, just check if the runner is contained in `supported_proton_runners`? Would it be an acceptable fix?

bottles/backend/runner.py
```diff
@@ -89,7 +89,7 @@ def runner_update(config: BottleConfig, manager: 'Manager', runner: str) -> Resu
         the host system. There are some exceptions, like the Soda and Wine-GE runners,
         which are built to work without the Steam Runtime.
         """
-        if "proton" in runner.lower() and RuntimeManager.get_runtimes("steam"):
+        if "proton" in runner.lower() and not runner.startswith("wine-ge-proton") and RuntimeManager.get_runtimes("steam"):
```
Kinsteen, line 92:
```suggestion
        if runner in manager.supported_proton_runners and RuntimeManager.get_runtimes("steam"):
```

Kinsteen:
@koplo199 It would be a perfect fix, even more so than what we had before. I tested quickly, and it seemed to work fine. More tests would be welcome, but I think it's the way to go.

SuperSamus:
Looking around the code, it seems that `"Proton" in runner` is used in multiple places, like [here](https://github.com/bottlesdevs/Bottles/blob/6954b8ad74f6ea68a5545fbe7d169f9f447760fb/bottles/backend/wine/winecommand.py#L419), or [here](https://github.com/bottlesdevs/Bottles/blob/6954b8ad74f6ea68a5545fbe7d169f9f447760fb/bottles/backend/wine/winecommand.py#L512). (Just grep `"proton"`/`"Proton"` to find them)
I guess these should be changed too, but I'm not familiar with the codebase (nor with Python), so I don't know how I should do `manager.supported_proton_runners` in these cases.

Kinsteen:
Okay. Would you be more comfortable closing this, and letting someone else do it, and change this like I suggested so we could merge? Someone else can do the other occurrences of this behavior, if you don't want to do it.

bottles/backend/managers/manager.py
```diff
@@ -345,17 +345,16 @@ def check_runners(self, install_latest: bool = True) -> bool:
...
-                f"{runner}lib/wine/x86_64-windows/winemenubuilder.exe",
-                f"{runner}lib32/wine/i386-windows/winemenubuilder.exe",
-                f"{runner}lib/wine/i386-windows/winemenubuilder.exe",
-            ]
-            for winemenubuilder in winemenubuilder_paths:
-                if winemenubuilder.startswith("Proton"):
-                    continue
-                if os.path.isfile(winemenubuilder):
-                    os.rename(winemenubuilder, f"{winemenubuilder}.lock")
+            if runner in self.supported_proton_runners:
```
Kinsteen, line 348:
```suggestion
            if runner not in self.supported_proton_runners:
```
If I'm not mistaken, we add this lock only if it's *not* a Proton runner

Kinsteen (changes requested):

PR #2597, opened 2023-01-21
backend: Better wait operation for async Repos

Kinsteen:
# Description
This PR replaces the Lock mechanism for repo operation with events, which are easier to use and better to maintain than Locks.

Fixes nothing, but makes life easier mostly

## Type of change
- [X] Backend change

# How Has This Been Tested?
- [X] locally

### DIFF
--- a/bottles/backend/managers/manager.py
+++ b/bottles/backend/managers/manager.py
@@ -269,7 +269,6 @@ class Manager:
     def organize_components(self):
         """Get components catalog and organizes into supported_ lists."""
         RepoStatus.repo_wait_operation("components.fetching")
-        RepoStatus.repo_start_operation("components.organizing")
         catalog = self.component_manager.fetch_catalog()
         if len(catalog) == 0:
             RepoStatus.repo_done_operation("components.organizing")
@@ -290,7 +289,6 @@ class Manager:
     def organize_dependencies(self):
         """Organizes dependencies into supported_dependencies."""
         RepoStatus.repo_wait_operation("dependencies.fetching")
-        RepoStatus.repo_start_operation("dependencies.organizing")
         catalog = self.dependency_manager.fetch_catalog()
         if len(catalog) == 0:
             RepoStatus.repo_done_operation("dependencies.organizing")
@@ -304,7 +302,6 @@ class Manager:
     def organize_installers(self):
         """Organizes installers into supported_installers."""
         RepoStatus.repo_wait_operation("installers.fetching")
-        RepoStatus.repo_start_operation("installers.organizing")
         catalog = self.installer_manager.fetch_catalog()
         if len(catalog) == 0:
             RepoStatus.repo_done_operation("installers.organizing")
--- a/bottles/backend/managers/manager.py
+++ b/bottles/backend/managers/manager.py
@@ -268,8 +268,8 @@ class Manager:
     @RunAsync.run_async
     def organize_components(self):
         """Get components catalog and organizes into supported_ lists."""
-        RepoStatus.repo_start_operation("components.organizing")
         RepoStatus.repo_wait_operation("components.fetching")
+        RepoStatus.repo_start_operation("components.organizing")
         catalog = self.component_manager.fetch_catalog()
         if len(catalog) == 0:
             RepoStatus.repo_done_operation("components.organizing")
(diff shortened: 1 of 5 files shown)


Kinsteen:
@StoneMoe I requested a review because I'm not sure if using a sleep is the best solution here, to wait the start of an operation. Do you have another idea?

Kinsteen:
I've reimplemented stuff using events instead of locks, which seems to be the best way to do this.

@StoneMoe suggestions:
- Can we use context manager instead of calling wait and done methods manually?
- Can we also use Enum for the lock names?

Those are valid questions that we will discuss. I would also like to add: can we generalize this class so it could be used somewhere else in the backend in the future, if we add more and more async code?

bottles/backend/repos/repo.py
```diff
@@ -88,24 +86,17 @@ def get_manifest(self, url: str, plain: bool = False):
...
     def repo_wait_operation(name: str):
-        lock = RepoStatus.LOCKS.setdefault(name, PyLock())
-        logging.debug(f"Wait operation {name}")
-        lock.acquire()
-        lock.release()
-        logging.debug(f"Done wait operation {name}")
\ No newline at end of file
+        event = RepoStatus.EVENTS.setdefault(name, PyEvent())
+        # By default when an Event is created, it will be false, so it will block
+        event.wait()
```
StoneMoe, line 101: I don't think they are idempotent
it will wait for event forever if the operation you wait ended too quick or never start
the execution order of async functions is hard to ensure

bottles/backend/managers/manager.py
```diff
@@ -303,7 +301,6 @@ def organize_dependencies(self):
     @RunAsync.run_async
     def organize_installers(self):
         """Organizes installers into supported_installers."""
-        RepoStatus.repo_start_operation("installers.organizing")
```
StoneMoe, line 306: Just cuirous, why removing these organizing lines?

bottles/frontend/views/bottle_dependencies.py
```diff
@@ -113,12 +113,11 @@ def new_dependency(dependency, plain=False):
...
             self.list_dependencies.append(entry)
 
-        def callback(result, error=False):
+        def callback(_result, _error=False):
             self.stack.set_visible_child_name("page_deps")
 
         def process_dependencies():
             time.sleep(.3)  # workaround for freezing bug on bottle load
-            RepoStatus.repo_wait_operation("dependencies.fetching")
             RepoStatus.repo_wait_operation("dependencies.organizing")
```
StoneMoe, line 121: should this be removed instead of removing wait for fetching?

bottles/backend/repos/repo.py
Kinsteen: I see what you mean. One solution could be to check if the operation existed and finished, and if yes, simply continuing? That would mean that this operation could only be "waited" once, let's say. It would work for the use case that we have, but then it's a problem when/if we generalize this class.

bottles/backend/managers/manager.py
Kinsteen: `start_operation` is not needed, only `wait` and `end`. When events are created, they are not set by default so it doesn't matter if we start or not the operation.

bottles/frontend/views/bottle_dependencies.py
Kinsteen: No, here we want to have the fetching and organizing done. As the organizing already depends on the fetching, it's not necessary. This was a safe-guard because of the old way I used locks that could create problems.

bottles/backend/repos/repo.py
StoneMoe: > One solution could be to check if the operation existed and finished

that's exactly what I mentioned before in Discord by "busy wait is the only option for this if you don’t want to introduce any other variable to record the fetching started state"

anyway this solution works for me :D

StoneMoe: My bad, the `wait` method won't wait forever, just ignore this comment
```
wait(timeout=None) will block until the internal flag is true. If the internal flag is true on entry, return immediately.
```

Kinsteen:
This will be reworked in #2628, so it is not needed anymore.

PR #2600, opened 2023-01-21
Update runtime automatically on bottles start

cyberphantom52:
# Description
This ensures that the non flatpak version of bottles always uses the latest available runtime. The core prefrences section has also been removed to make the ui consistent between flatpak and non flatpak versions.

Fixes Nothing

## Type of change

# How Has This Been Tested?
- [X] Locally

### DIFF
--- a/bottles/backend/managers/manager.py
+++ b/bottles/backend/managers/manager.py
@@ -1416,6 +1416,7 @@ class Manager:
         subprocess.run(["rm", "-rf", path], stdout=subprocess.DEVNULL)
 
         del self.local_bottles[config.Name]
+        self.update_bottles(silent=True)
 
         logging.info(f"Deleted the bottle in: {path}")
         return True
--- a/bottles/backend/managers/manager.py
+++ b/bottles/backend/managers/manager.py
@@ -1415,11 +1415,7 @@ class Manager:
         path = ManagerUtils.get_bottle_path(config)
         subprocess.run(["rm", "-rf", path], stdout=subprocess.DEVNULL)
 
-        local_bottles_tmp = self.local_bottles.copy()
-        for b in local_bottles_tmp.values():
-            if b.Path == config.Path:
-                del self.local_bottles[b.Name]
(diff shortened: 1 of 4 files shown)


bottles/backend/managers/manager.py
```diff
@@ -410,35 +410,35 @@ def check_runners(self, install_latest: bool = True) -> bool:
...
-                except StopIteration:
-                    return False
-            return False
-
-        runtime = runtimes[0]  # runtimes cannot be more than one
-        manifest = os.path.join(Paths.runtimes, runtime, "manifest.yml")
+            logging.warning("No runtime found.")
+            install_latest = True
+        else:
+            runtime = runtimes[0]  # runtimes cannot be more than one
```
Kinsteen, line 427: Are you sure? If users have 0.5 installed, then 0.6 will be installed. But then, both will be in the folder? There is nothing that deletes old version, unless I'm mistaken?

bottles/backend/managers/manager.py
```diff
@@ -410,35 +410,35 @@ def check_runners(self, install_latest: bool = True) -> bool:
         self.runners_available = sorted(self.runners_available, reverse=True)
         return True
 
+    @RunAsync.run_async
     def check_runtimes(self, install_latest: bool = True) -> bool:
-        self.runtimes_available = []
+        RepoStatus.repo_wait_operation("components.fetching")
+        RepoStatus.repo_wait_operation("components.organizing")
         if "FLATPAK_ID" in os.environ:
```
Kinsteen, line 417: This block should be removed for Flatpak, as we're using the components version

cyberphantom52:
@Kinsteen i guess i should just close this PR and work on shipping EAC and BE as components?

Kinsteen:
Yes, I think that would be good!

cyberphantom52:
We decided that it was best to not ship runtime in the first place and rather switch to only shipping EAC and BattleEye runtimes.

PR #2665, opened 2023-02-09
Added event manager + tests

Kinsteen:
# Description
This adds the EventManager class + some events, and tests about it.

Based on the work here: https://github.com/bottlesdevs/Bottles/pull/2597

Fixes #(issue)

## Type of change
- [x] Backend rewrite

# How Has This Been Tested?
- [x] locally

### DIFF
--- a/bottles/backend/repos/repo.py
+++ b/bottles/backend/repos/repo.py
@@ -25,6 +25,7 @@ from bottles.backend.logger import Logger
 from bottles.backend.utils import yaml
 from bottles.backend.utils.threading import RunAsync
 from bottles.frontend.utils.gtk import GtkUtils
+from bottles.backend.state import EventManager, Events
 
 logging = Logger()
 
@@ -39,14 +40,12 @@ class Repo:
         @GtkUtils.run_in_main_loop
         def set_catalog(result, error=None):
             self.catalog = result
-            RepoStatus.repo_done_operation(self.name + ".fetching")
+            EventManager.done(Events(self.name + ".fetching"))
             if callback: callback()
 
         RunAsync(self.__get_catalog, callback=set_catalog, index=index, offline=offline)
 
     def __get_catalog(self, index: str, offline: bool = False):
-        RepoStatus.repo_start_operation(self.name + ".fetching")
-
         if index in ["", None] or offline:
             return {}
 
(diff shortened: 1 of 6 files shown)


bottles/tests/state/test_events.py
```diff
@@ -0,0 +1,85 @@
+"""EventManager tests"""
```
StoneMoe, line 1: Can we have more comments about how these tests works?
the timing conditions looks scary and not robust enough to me

bottles/backend/repos/repo.py
```diff
@@ -39,14 +40,12 @@ def __init__(self, url: str, index: str, offline: bool = False, callback=None):
         @GtkUtils.run_in_main_loop
         def set_catalog(result, error=None):
             self.catalog = result
-            RepoStatus.repo_done_operation(self.name + ".fetching")
+            EventManager.done(Events(self.name + ".fetching"))
```
StoneMoe, line 43: it's easy to make mistake in the future because of overlooking during making changes.
is there any alternative way to do this in a more explicit way, or can we at least add some guard code somewhere in `EventManager` to make sure we are crafting Enum object with correct names?

bottles/tests/state/test_events.py
Kinsteen: I'm not sure what is the best way to test that events work in this case, with something other than just timing different threads. As per my testing, it seems robust *enough*, but if you have ideas on how to test, I'll be happy to hear them!

bottles/backend/repos/repo.py
Kinsteen: If we try to craft an event that doesn't have a valid value, there will be a very clear error, so I think it's okay to let it like this:
```
>>> from enum import Enum
>>> class TestEnum(Enum):
...   Test1 = "test1"
...   Test2 = "test2"
... 
>>> TestEnum("test3")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.10/enum.py", line 385, in __call__
    return cls.__new__(cls, value)
  File "/usr/lib/python3.10/enum.py", line 710, in __new__
    raise ve_exc
ValueError: 'test3' is not a valid TestEnum
```

I'm not aware of better ways to do it as this code is very generic for good reasons, so imo it's okay, what do you think?

bottles/backend/state.py
```diff
@@ -131,8 +136,31 @@ def get_lock(cls, name: Locks) -> PyLock:
 
 
 class EventManager:
+    """
+    This class manages events, which are one-time events (can be reset) during the lifecycle of the app.
+    You can wait for the event to occur, or set it when the associated operations is finished.
```
Kinsteen, line 141:
```suggestion
    You can wait for the event to occur, or set it when the associated operations are finished.
```

bottles/backend/repos/repo.py
StoneMoe: looks good to me if it's already throwing error with invalid names

bottles/tests/state/test_events.py
StoneMoe: I think we should test against more specific behavior instead of undistinguished call results.

for example, we should have tests like:
- when we call `wait()` multiple times with same name
  - should only create one `PyEvent` in `EventManager` to make sure they are referencing the same event object
- when we call`done()` or `reset()`
  - the internal flag in the corresponding `PyEvent` should be correctly set
- when we call `done()` or `reset()` with same name in multiple function
  - they should using the same `PyEvent` since the registry in `EventManager` should be singleton

StoneMoe (approved):
Looks great now!

PR #2684, opened 2023-02-17
Make unspecified 'Arch' to be listed for 32-bit and 64-bit architecture

### LINKED ISSUE #2166: [Bug]: On New Custom 32bit Bottle - it doesn't list out the Dependencies.
### Describe the bug

Hi, 

when I try to create a New Custom 32bit Bottle, once the bottle is created, if I go to Dependencies, the list of `.dlls` to install does not appear. 

In the 64bit bottles, the Dependencies can be seen and can be easily installed.

### To Reproduce

1. Create New Bottle
2. Choose: Custom
3. Architecture 32 bit
4. Enter into the New Created 32 bit Bottle
5. On list to the left - Go to the voice `Dependencies`
6. And no one `.dll` list is showing.

### Package

Flatpak from Flathub

### Distribution

Ubuntu 22.04.1 LTS

### Troubleshooting Logs

```shell
Version: 2022.10.14.1
Display:
    X.org: true
    X.org (port): :1
    Wayland: false
Graphics:
    vendors:
        nvidia: &amp;id002
            vendor: nvidia
            envs:
                __NV_PRIME_RENDER_OFFLOAD: '1'
                __GLX_VENDOR_LIBRARY_NAME: nvidia
                __VK_LAYER_NV_optimus: NVIDIA_only
            icd: /usr/lib/x86_64-linux-gnu/GL/vulkan/icd.d/nvidia_icd.json:/usr/lib/i386-linux-gnu/GL/vulkan/icd.d/nvidia_icd.json
        intel: &amp;id001
            vendor: intel
            envs:
                DRI_PRIME: '1'
            icd: /usr/lib/x86_64-linux-gnu/GL/vulkan/icd.d/intel_icd.x86_64.json:/usr/lib/i386-linux-gnu/GL/vulkan/icd.d/intel_icd.i686.json
    prime:
        integrated: *id001
        discrete: *id002
Kernel:
    Type: Linux
    Version: 6.0.1-surface
Distro:
    Name: GNOME
    Version: '43'
Disk:
    Total: 16748654592
    Free: 16748494848
RAM:
    MemTotal: "33.5\_GB"
    MemAvailable: "26.8\_GB"
Bottles_envs: null
```

### Display Protocol

X11

I don't know

### Additional context

_No response_

Zylquinal:
# Description
Make unspecified 'Arch' to be listed for 32-bit and 64-bit architecture. As we know that nearly all of the dependencies listed on the repository doesn't have 'Arch' attribute, and with the missing of these attributes, nearly all of the dependencies listed on the repo can't be seen on the 32-bit bottle environment. It happen due to the way the code return 'Win64' on null or missing attribute.

Fixes #2166

## Type of change

# How Has This Been Tested?
- [x] Local Machine

[image of installed dependencies on 32-bit environment](https://imgur.com/ZWPlb6J)
[image of installed program using the modified bottle](https://imgur.com/a/82yNkeU)

### DIFF
--- a/bottles/backend/managers/dependency.py
+++ b/bottles/backend/managers/dependency.py
@@ -139,12 +139,15 @@ class DependencyManager:
                     _res = self.install(config, [_ext_dep, _dep])
                     if not _res.status:
                         return _res
-
         for step in manifest.get("Steps"):
             """
             Here we execute all steps in the manifest.
             Steps are the actions performed to install the dependency.
             """
+            arch = step.get("for", "win64_win32")
+            if config.Arch not in arch:
+                continue
+
             res = self.__perform_steps(config, step)
             if not res.status:
                 GLib.idle_add(self.__operation_manager.remove_task, task_id)
--- a/bottles/frontend/ui/details-dependencies.blp
+++ b/bottles/frontend/ui/details-dependencies.blp
@@ -81,6 +81,14 @@ Popover pop_context {
       tooltip-text: _("Read Documentation.");
       text: _("Documentation");
     }
+    .GtkModelButton btn_show_all {
+      tooltip-text: _("Show all dependencies.");
+      text: _("Show All Dependencies");
+    }
+    .GtkModelButton btn_show_allowed {
+      tooltip-text: _("Show allowed dependencies.");
+      text: _("Show Allowed Dependencies");
+    }
   }
 }
 
--- a/bottles/frontend/views/bottle_dependencies.py
+++ b/bottles/frontend/views/bottle_dependencies.py
@@ -129,7 +129,8 @@ class DependenciesView(Adw.Bin):
                     if dep[0] in self.config.Installed_Dependencies:
                         continue  # Do not list already installed dependencies'
 
-                    if dep[1].get("Arch", "win64") != self.config.Arch:
+                    dependency_arch = dep[1].get("Arch")
+                    if dependency_arch is not None and dependency_arch != self.config.get("Arch"):
                         # NOTE: avoid listing dependencies not supported by the bottle arch
                         continue
 
--- a/bottles/frontend/views/bottle_dependencies.py
+++ b/bottles/frontend/views/bottle_dependencies.py
@@ -14,7 +14,6 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
-
 import time
 from typing import Optional
 
@@ -35,6 +34,8 @@ class DependenciesView(Adw.Bin):
     # region Widgets
     list_dependencies = Gtk.Template.Child()
     btn_report = Gtk.Template.Child()
+    btn_show_all = Gtk.Template.Child()
+    btn_show_allowed = Gtk.Template.Child()
     btn_help = Gtk.Template.Child()
     entry_search = Gtk.Template.Child()
     actions = Gtk.Template.Child()
@@ -53,6 +54,7 @@ class DependenciesView(Adw.Bin):
         self.manager = details.window.manager
         self.config = config
         self.queue = details.queue
+        self.show_all = False
 
         self.ev_controller.connect("key-released", self.__search_dependencies)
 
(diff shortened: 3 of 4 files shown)


Kinsteen:
I'm afraid this will cause problems when installing some dependencies, as they were made for 64 bits... For example, vcredist2015 installs both x86 and x64 version ; does it crash? Does it silently fail?

Zylquinal:
That's indeed true, but i guess it should be done at repository level. Instead of making a fallback to `win64`. i think all dependencies must have `Arch` attribute. Making it clearer for people to know which dependency that actually support the architecture.

Zylquinal:
Since some dependencies could support both architectures, i think the dependency could look like this instead.
```yaml
Arch: [win32, win64]
Steps: 
- action: install_exe
  arch:
    win32:
      file_name: VC_redist.x86.exe
      url: https://download.microsoft.com/download/9/3/F/93FCF1E7-E6A4-478B-96E7-D4B285925B00/vc_redist.x86.exe
      rename: vcredist2015_x86.exe
      file_checksum: 1a15e6606bac9647e7ad3caa543377cf
      file_size: 13767776
      arguments: /quiet /norestart
    win64:
      file_name: VC_redist.x64.exe
      url: https://download.microsoft.com/download/9/3/F/93FCF1E7-E6A4-478B-96E7-D4B285925B00/vc_redist.x64.exe
      rename: vcredist2015_x64.exe
      file_checksum: 27b141aacc2777a82bb3fa9f6e5e5c1c
      file_size: 14572000
      arguments: /quiet /norestart
 ```

**Comment by Kinsteen** [comment]
I think it would be best to have an option for steps like:
```yaml
for:
  - "x86"
 ```
that you can specify for each step. It could be an array, assumed per default to be `["x86", "x64"]`, so that we can specify the architecture for each step.

Most of 32 bits stuff also needs to be installed for 64 bits bottles, and it would require less changes on Bottles/dependencies side. The solution you proposed seemed a bit too much imo

**Comment by Zylquinal** [comment]
Well i guess that's true that it is indeed too much, especially when deciding on updating the repo structure. But what about this.

[image 1](https://imgur.com/YEb1A7I)
[image 2](https://imgur.com/zoay8Rn)

Instead of changing the check, user could press the button to show all of the dependencies, while slowly working on which step should be added on x86. Since i guess it would take quite a lot of time testing each lib and updating each repo step. And anyway i new to this project, never thought it would be like this when trying to fix the custom environment not showing any lib on my Bottles.

**Comment by Kinsteen** [comment]
This "bug" is there since a long time because there's no easy fix unfortunately...

**Comment by Zylquinal** [comment]
Just changed a few thing [here](https://github.com/Zylquinal/Bottles/tree/dep-fix), with the config [example](https://pastebin.com/hPMa6syb). The program will retain its compatibility with older version, and it should be fine for older version to safely read the config. On the `Arch`, i decide to make it like `win64`, `win32`, `win32_win64` (or reversed). It also added a button on the dependency menu, so people could choose to view every dependencies without any architecture filter. What do you think?

**Comment by Kinsteen** [comment]
I would much prefer to have manifests like this:
```yaml
Name: vcredist2015
Description: Microsoft Visual C++ Redistributable for Visual Studio 2015
Provider: Microsoft
License: Microsoft EULA
License_url: https://www.microsoft.com/web/webpi/eula/net_library_eula_enu.htm
Dependencies: []
Steps:
- action: install_exe
  file_name: VC_redist.x86.exe
  url: https://download.microsoft.com/download/9/3/F/93FCF1E7-E6A4-478B-96E7-D4B285925B00/vc_redist.x86.exe
  rename: vcredist2015_x86.exe
  file_checksum: 1a15e6606bac9647e7ad3caa543377cf
  file_size: 13767776
  arguments: /quiet /norestart

- action: install_exe
  file_name: VC_redist.x64.exe
  url: https://download.microsoft.com/download/9/3/F/93FCF1E7-E6A4-478B-96E7-D4B285925B00/vc_redist.x64.exe
  rename: vcredist2015_x64.exe
  file_checksum: 27b141aacc2777a82bb3fa9f6e5e5c1c
  file_size: 14572000
  arguments: /quiet /norestart
  for:
    - win64
  
- action: override_dll
  dll: api-ms-win-crt-private-l1-1-0
  type: native,builtin
```
The less changes the better, and even in the Python code it would not require much changes. We assume that by default, each dependency is 32 and 64 bits compatible, and same for each step, unless there is a `for` clause like in the thing I sent. I think it's better, but I would also like thoughts from other devs

Zylquinal:
Gosh just realized what you mean, that's indeed better XD. Guess thing could go bad when you haven't have any sleep for 24+ hours. Just changed it like what you're suggesting. 

This is the [example](https://pastebin.com/UCu3ciC6) manifest, and [this](https://pastebin.com/MGS1iBQF) is how it successfully run on 64-bit environment. It also will not affect any older version of the manifest, [example of installing dependency with older version of the manifest](https://pastebin.com/qRPwtUqM). I also change it to return `win64_win32` when there's no `Arch` detected on the manifest, and also making the one without `for` as the `global` action.

TheEvilSkeleton:
I'm marking as draft, as it's still under discussion.

orowith2os:
Closing, redirecting to https://github.com/bottlesdevs/Bottles/pull/2861, which someone needs to pick up.

orowith2os:
This was done in https://github.com/bottlesdevs/Bottles/pull/3101

PR #2747, opened 2023-02-26
Initial mypy support

### LINKED ISSUE #2746: Use mypy to check python code
Every python project I've co-developed had mypy in the CI, which has prevented a ton of problems, so I think this project should also have support enabled.

After installing mypy with missing types - `mypy --install-types`, mypy produced 180 errors in code e.g.
```
bottles/backend/repos/component.py:26: error: Unsupported right operand type for in ("None")
bottles/backend/repos/component.py:27: error: Value of type "None" is not indexable
bottles/frontend/views/bottle_dependencies.py:34: error: Need type annotation for "__registry" (hint: "__registry: List[<type>] = ...")
bottles/backend/managers/steam.py:248: error: Item "None" of "Optional[BottleConfig]" has no attribute "Parameters"
bottles/backend/managers/steam.py:356: error: Invalid type comment or annotation
bottles/backend/managers/steam.py:383: error: Incompatible types in assignment (expression has type "List[str]", variable has type "str")
bottles/backend/utils/manager.py:219: error: Argument 3 to "extract_icon" of "ManagerUtils" has incompatible type "Optional[Any]"; expected "str"
bottles/backend/utils/manager.py:245: error: Argument 1 to "quote" has incompatible type "Optional[Any]"; expected "str"
bottles/backend/utils/manager.py:300: error: Argument "custom_path" to "open_filemanager" of "ManagerUtils" has incompatible type "Optional[Any]"; expected "str"
bottles/backend/wine/winecommand.py:88: error: Incompatible default for argument "arguments" (default has type "bool", argument has type "str")
bottles/backend/wine/winecommand.py:89: error: Incompatible default for argument "environment" (default has type "bool", argument has type "Dict[Any, Any]")
bottles/backend/wine/winecommand.py:444: error: Incompatible types in assignment (expression has type "Optional[str]", variable has type "str")
```

Looks that there is really big amount of valid errors - like e.g. variables randomly changing its type:
File list.py
```python
        '''Format update date'''
        update_date = _("N/A")    # Here variables is str
        if self.config.Update_Date:
            try:
                update_date = datetime.strptime(self.config.Update_Date, "%Y-%m-%d %H:%M:%S.%f")    # Here variables is datetime
                update_date = update_date.strftime("%d %B, %Y %H:%M:%S")   # and boom, here is again str
            except ValueError:
                update_date = _("N/A")
```

qarmin:
# Description
This PR adds initial mypy support.
I only fixed some occurrences of `implictic-optional` errors.

For now there is still a lot of disabled checks in `mypy.ini`, that needs to be enabled, but this should be done incrementally in several small PR's.
 
Fixes #2746

First CI run should fail because I commented disabled check in mypy.ini, to see if CI works properly

## Type of change

## How Has This Been Tested?
- [x] - tested locally


.github/workflows/mypy.yml
```diff
@@ -0,0 +1,33 @@
+on:
+  pull_request:
+name: Mypy
+jobs:
+  pylint-checker:
```
Kinsteen, line 5: This should be renamed

.github/workflows/mypy.yml
```diff
@@ -0,0 +1,33 @@
...
+      run: |
+        python -m pip install --upgrade pip
+        pip install wheel
+        pip install mypy
+        pip install -r requirements.txt
+        pip install pycairo
+        pip install PyGObject
+        yes | mypy --install-types
+    - name: Run mypy analysis
+      id: pylint-result
```
Kinsteen, line 30: This should be removed if not used, renamed if used later

Kinsteen (commented):
I don't have any experience with MyPy, but it seems like it would be a great addition! However, it would be best if it could also use a generic version of `utils/pylint-parser.py`, so it could run only on modified files in a PR, instead of all the files, which would be pretty much unreadable. It would also be nice if there could be another action that could comment the PR with the results, like with Pylint again (maybe in the same comment, to avoid too many comments?)

qarmin:
> However, it would be best if it could also use a generic version of utils/pylint-parser.py, so it could run only on modified files in a PR, instead of all the files, which would be pretty much unreadable. 

This is very bad idea,
Mypy greatest strength is to show type(and others) errors in files that were not directly changed e.g. 
A.py file
```
from C import func

func("A", 1)
```
C.py file
```
def func(st: str, in: int):
    pass
```
Swapping arguments in C file will automatically show error in A file(if checking only the changed files was done, then basically there would be no point in using mypy)

> so it could run only on modified files in a PR, instead of all the files, which would be pretty much unreadable

In the main branch, mypy should always detect 0 errors, so if in any PR it indicates more than 0 it means that the person who created it must correct them and no errors should be hidden from it(in my experience, I know that very rarely is the list of errors greater than 5/10)

Kinsteen:
I see, I was saying that it could only log for files that were modified, but run on the whole codebase. It will take a bit of time to fix all types errors, but it's interesting to have the log only on modified files in the PR, don't you think? That's what is currently done with Pylint for example, as otherwise the logs would be thousands of lines long

StoneMoe:
Great to see any progress about Code Quality improvement!
which is we are focusing in recent versions.

however there are two issues I found that need to be fix in this PR:
1. Please use `Optional[type]` instead of `type | None` to match our codebase
2. when a argument is like`param: Type = None`, it doesn't means you can simply change it to `param: Optional[Type] = None`, the default value `None` can be an indicator that the argument is not passed in, which doen't means caller can explicit pass a `None` in.

Also, is MyPy our only choice?
there are serveral type-aware static checkers out there, like Mypy, Pytype, Pyright, and Pyre.

qarmin:
> when a argument is likeparam: Type = None, it doesn't means you can simply change it to param: Optional[Type] = None, the default value None can be an indicator that the argument is not passed in, which doen't means caller can explicit pass a None in.

For me this `: Type = None` looks counter intuitive, this `is no longer the recommended behavior`(by mypy) and also mypy converts this internally to `: Optional[Type] = None`. I don't see too big advantages of current behavior
https://github.com/python/peps/pull/689/files

StoneMoe:
> > when a argument is likeparam: Type = None, it doesn't means you can simply change it to param: Optional[Type] = None, the default value None can be an indicator that the argument is not passed in, which doen't means caller can explicit pass a None in.
> 
> For me this `: Type = None` looks counter intuitive, this `is no longer the recommended behavior`(by mypy) and also mypy converts this internally to `: Optional[Type] = None`. I don't see too big advantages of current behavior https://github.com/python/peps/pull/689/files

I think the sentence `is no longer the recommended behavior` is describing that `Type checkers should move
towards requiring the optional type to be made explicit` instead of `implicitly making an argument Optional`.

I'm bad in English, correct me if I'm wrong.

StoneMoe:
> For me this `: Type = None` looks counter intuitive
I'm agree with this to make everything less ambiguous tho.

qarmin:
Probably fixed CI by installing manually all typing libraries.
Mypy takes 7s(<1s with cache) to scan project on my OS, so I don't think that would be big gain in adding cache to CI

Kinsteen:
Seems to have worked, but your update to requirements broke Pytest action. Can you fix it? Thanks!

Is it normal that we don't have any errors?

When all of it is done, I think we can merge! I'll maybe add a commenter so we can have a quick link to the report, but I'll do it later

PR #2772, opened 2023-03-04
Add ruff as linter

qarmin:
# Description
This PR add as linter [ruff](https://github.com/charliermarsh/ruff) - alternative to pylint, flake8 etc.

This commit only adds CI job and config file with suppressed every error found by this tool.
Fixes(removing one by one ignored errors from toml file) should be applied incrementally in other PR's

## But why ruff?
- CI Problem - currently pylint emits for this repository ~3000 lines of errors, that no-one wants to fix. This number is greatly decreased, by only showing problems in changed files in PR,  but still I don't think that ever this number will be 0. Ruff in opposite uses simple ignore list(current pylint configuration file is quite complicated - https://github.com/bottlesdevs/Bottles/blob/main/.pylintrc), and aims to show exactly 0 errors, so it can be forced in CI.
- Performance - this is really amazing, pylint this entire repo scanned in 40 seconds
```
real	0m40,991s
user	0m40,399s
sys	0m0,509s
```
but ruff only in ~0.06 second
```
real	0m0,062s
user	0m0,151s
sys	0m0,027s
```
- Automation - many ruff rules offers automatic fixes by typing `ruff . --fix`

This PR not touch pylint, but I think, that someday pylint checks should be removed from repo.

EDIT:
I forgot to write about disadvantages:
- app not have yet stable release - ~I tested ~1000 most popular crates, and app crashed only twice, so for me is quite stable~ - already all crashes which I found were fixed
- not all rules are implemented - ruff tries to be all in one solution, that combines rules from multiple apps, but due limits not all are available yet

## Type of change

# How Has This Been Tested?
- [x] Test by running `ruff .`

### DIFF
--- a/.github/workflows/ruff-checker.yml
+++ b/.github/workflows/ruff-checker.yml
@@ -1,0 +1,21 @@
+on:
+  pull_request:
+name: Ruff - checker
+jobs:
+  ruff-checker:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - name: Set up Python 3.10
+      uses: actions/setup-python@v3
+      with:
+        python-version: "3.10"
+    - name: Install Python dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install ruff==0.0.254
+    - name: Analysing the code with ruff
+      continue-on-error: true
+      run: ruff .
--- a/requirements.dev.txt
+++ b/requirements.dev.txt
@@ -11,3 +11,4 @@ types_pycurl
 types_requests
 types_docutils
 pylint
+ruff==0.0.285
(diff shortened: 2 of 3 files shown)


StoneMoe:
To be honest, I don't see a strong enough motivation to switch from pylint to ruff. Moreover, it would be better if we compared other similar tools before making a final decision.

Since Pylint was initially brought into the project by @Kinsteen, it would be beneficial for us to hear his thoughts and ideas before making any further changes or decisions on this matter.

Kinsteen:
I don't have much opinion myself. I like the fact that it's faster though

We don't have a very specific use of Pylint, so maybe it will do as well. Not sure how to go forward with this

mirkobrombin:
As far as I know, pylint has a lot more rules but it's also a lot slower. I believe speed is not a benefit factor here.

qarmin:
The problem is that pylint have a lot of false positives and pylint style is not forced in any way in this repo(all errors can be easily ignored, when merging PR)

Looks that pylint supports ~400 rules, but ruff ~500(115 rules pylint rules are reimplemented - https://github.com/charliermarsh/ruff/issues/970)

Not sure if any from non implemented rules are important.

Even pylint repo use ruff - https://github.com/PyCQA/pylint/commit/fbebb952dd946ab9122448f334e97d39b5b41317

mirkobrombin:
Good point

PR #2823, opened 2023-03-27
po: update POTFILES, PO files; make translator credits a translatable string

kbdharun:
# Description

- Updated POTFILES as `connection.py` component was removed.
- Updated POT files and PO Files.
- Currently, translator credits are maintained manually which is tedious to maintain (and it hasn't been maintained), I propose making it a translatable term instead so the credits for translators are shown when people are accessing it in the respective languages.

## Type of change

# How Has This Been Tested?
- [x] Only POTFILES was manually touched, translation files are able to be updated with Meson using the following commands:

```bash
meson setup /tmp/translation-build
meson compile -C /tmp/translation-build bottles-pot
meson compile -C /tmp/translation-build bottles-update-po
```


TheEvilSkeleton:
> Currently, translator credits are maintained manually which is tedious to maintain (and it hasn't been maintained), I propose making it a translatable term instead so the credits for translators are shown when people are accessing it in the respective languages.

You have my support :+1:

PR #2824, opened 2023-03-27
Add support for newer vkd3d-proton + bug fix

koplo199:
# Description

Newer vkd3d-proton will include a new dll : `d3d12core.dll`.
The `d3d12.dll` now being a wrapper around this new dll, reference: https://github.com/HansKristian-Work/vkd3d-proton/commit/fbf03829ebaaf068f9166c668a7f1c9265999915

Additionally, I found why `del found[path][dll]` was throwing a `TypeError`: `found[path]` is a list not a dict, so we must use 
`found[path].remove(dll)` instead.

## Type of change

# How Has This Been Tested?
Tested locally using https://github.com/bottlesdevs/components/pull/236 allowing to use an up-to-date `vkd3d-proton` including the new dll.

### DIFF
--- a/bottles/backend/dlls/dll.py
+++ b/bottles/backend/dlls/dll.py
@@ -20,6 +20,7 @@ import shutil
 from glob import glob
 from typing import NewType
 from abc import abstractmethod
+from copy import deepcopy
 
 from bottles.backend.logger import Logger
 from bottles.backend.models.config import BottleConfig
@@ -34,6 +35,7 @@ logging = Logger()
 class DLLComponent:
     base_path: str = None
     dlls: dict = {}
+    checked_dlls: dict = {}
     version: str = None
 
     def __init__(self, version: str):
@@ -47,7 +49,7 @@ class DLLComponent:
         pass
 
     def check(self):
-        found = self.dlls.copy()
+        found = deepcopy(self.dlls)
 
         if None in self.dlls:
             logging.error(f"DLL(s) \"{self.dlls[None]}\" path haven't been found, ignoring...")
@@ -66,7 +68,7 @@ class DLLComponent:
         if len(found) == 0:
             return False
 
-        self.dlls = found
+        self.checked_dlls = found
         return True
 
     def install(self, config: BottleConfig, overrides_only: bool = False, exclude=None):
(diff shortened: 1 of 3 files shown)


koplo199:
> I'm pretty sure this will break old versions of VKD3D that doesn't have two DLLs...



More seriously @Kinsteen you're probably right, but I'm still 99.9% sure it will work alright: from what I understand if a dll is missing then it is simply not used. The bug prevented this correct behavior by deleting the whole folder if one of the dll went missing, but it is now fixed

Kinsteen:
In any case we *must* test before we merge

koplo199:
Yes, I really should set a dev environment for Bottles, but I don't want to mess with my local install: any tip for setting up an isolated dev environment? I looked at the doc but couldn't find any info on that matter

TheEvilSkeleton:
You could use the Flatpak artifacts. However, I suggest to backup your existing Bottles flatpak files, which are located in `~/.var/app/com.usebottles.bottles`. If you want, I can send you the direct link to the artifact that you can download

koplo199:
> You could use the Flatpak artifacts. However, I suggest to backup your existing Bottles flatpak files, which are located in ~/.var/app/com.usebottles.bottles. If you want, I can send you the direct link to the artifact that you can download

Unfortunately, I can't backup my existing files: there are multiple gigabytes of it and I don't have space left on my disk (hence why I asked if there was a way to set up an isolated dev environment).

I guess I could set up an additional user on my system, and then logging to it when developing/testing, but it is not very convenient (need to migrate existing config, etc).

A way to use another folder than `~/.var/app/com.usebottles.bottles` like `~/.var/app/com.usebottles.bottles-devel` or so would be perfect. Is there any way to achieve that? By editing a config maybe?

mirkobrombin:
> A way to use another folder than ~/.var/app/com.usebottles.bottles like ~/.var/app/com.usebottles.bottles-devel or so would be perfect. Is there any way to achieve that? By editing a build config maybe?

You can edit local source of Bottles to point to a different share path, should be under the globals.py file.

koplo199:
> You can edit local source of Bottles to point to a different share path, should be under the globals.py file.

Nice, thanks for the hint. Would you accept a PR that does that automatically when a `devel` build is detected?

mirkobrombin:
I'm not sure that's something you would expect, we may have to explain this

koplo199:
I don't know, imho it makes sense: a `dev` environment should be as clean as possible. That way, we make sure to replicate a first time use, thus avoiding any problem with caching (dependencies, components, old config file no longer shipped by default, or containing different settings, etc).

I'll investigate for personal use, we'll see

koplo199:
I ended up changing:
https://github.com/bottlesdevs/Bottles/blob/8bb19bbe84893a012fbc51693930a3a36f85414f/com.usebottles.bottles.yml#L1
to `id: com.usebottles.bottles.Devel` to achieve a separated Devel environment. I don't know if a manifest could be included in another, but creating a `com.usebottles.bottles.Devel.yml` including `com.usebottles.bottles.yml` and just changing its `id` would be the upstream-able solution.

Anyway, it allowed me to test and fix a bug which happens when switching back and forth between two versions not containing the same dlls.

After someone else can confirm it also works correctly on its system, It should be ready to merge.

Issue #2755, opened 2023-02-28
FSR is enabled on both Wine and Gamescope (instead of just one of them)


SuperSamus:
When FSR is enabled on Bottles' settings, it sets environment variables and arguments for both [Wine](https://github.com/bottlesdevs/Bottles/blob/6e583e408e18a47ffdcf83c46443b11badbdc22a/bottles/backend/wine/winecommand.py#L349) and [Gamescope](https://github.com/bottlesdevs/Bottles/blob/6e583e408e18a47ffdcf83c46443b11badbdc22a/bottles/backend/wine/winecommand.py#L557), which means that if you use a Wine version that supports FSR (like GE) and Gamescope, you are getting two separate instances of FSR, which is not how it's supposed to be used.
In theory, you could simply give priority to Gamescope's FSR, since there is no need to check whether the Wine version supports it. However, Wine-GE enables FSR by default, because due to the nature of the hack, disabling it causes problems (related: #2636). Thus, the right logic should be "if both FSR and Gamescope are enabled, don't enable it for Gamescope if the Wine version supports FSR".

There's more: some settings conflict with each other, like FSR's "Quality Mode".
This is a Wine-exclusive feature (I'm implying Wine-GE), set with `WINE_FULLSCREEN_FSR_MODE`: what it does is forcing a different set of resolutions of a game, plus the native one. It has two problems:
- If you use a Wine version that doesn't support FSR, and instead use it via Gamescope, then this setting does nothing, and Bottles doesn't communicate it.
- If you instead use a Wine version that supports FSR, it conflicts with Gamescope's game resolution: setting both of these will crash the game at the start[^1].
Bottles' UI should clearly communicate what is enabled given the other settings and what is not. How should this be done (for this specific example)?

So, solutions?
- Just remove FSR's "Quality Mode", and get the problem out of the way.
- Don't care if a setting is silently doing nothing.
- Reorganizing somehow the settings so that what is enabled and what is not is clearly communicated.

So, that's why I'm opening the issue.

(By the way, Wine has `WINE_FULLSCREEN_FSR_CUSTOM_MODE` (not implemented in Bottles), which adds a custom resolution in the resolution list. This however does not conflict with Gamescope.)

[...]

Issue #2636, opened 2023-01-30
[Bug]: FSR enabled in Wine-GE when disabled in Bottles


emansom:
### Describe the bug

Wine-GE _[always](https://github.com/GloriousEggroll/wine-ge-custom/blob/GE-Proton7-36/patches/proton/71-invert-fsr-logic.patch)_ enables FSR unless the environment variable `WINE_FULLSCREEN_FSR` equals `0`.

### To Reproduce

1. Configure Bottles to use Wine-GE
2. Launch a game
3. Observe it applying a slight sharpened filter (FSR)

### Package

Flatpak from Flathub

### Distribution

Arch Linux

### Debugging Information

```shell
Official Package: false
Version: 50.2-6f2d8635
DE/WM: gnome
Display:
    X.org: true
    X.org (port): :0
    Wayland: true
Graphics:
    vendors:
        amd: &amp;id002
            vendor: amd
            envs:
                DRI_PRIME: '1'
            icd: /usr/share/vulkan/icd.d/amd_pro_icd64.json:/usr/share/vulkan/icd.d/radeon_icd.x86_64.json:/usr/share/vulkan/icd.d/radeon_icd.i686.json
        intel: &amp;id001
            vendor: intel
            envs:
                DRI_PRIME: '1'
            icd: /usr/share/vulkan/icd.d/intel_hasvk_icd.x86_64.json:/usr/share/vulkan/icd.d/intel_icd.x86_64.json:/usr/share/vulkan/icd.d/intel_hasvk_icd.i686.json:/usr/share/vulkan/icd.d/intel_icd.i686.json
    prime:
        integrated: *id001
        discrete: *id002
Kernel:
    Type: Linux
    Version: 6.1.8-zen1-1-zen
Disk:
    Total: 118888927232
    Free: 30046064640
RAM:
    MemTotal: 31.2GiB
    MemAvailable: 22.4GiB
Bottles_envs: null
Tools and Libraries:
    cabextract: true
    p7zip: true
    patool: true
    glibc_min: '2.36'
    icoextract: true
    pefile: true
    orjson: true
    markdown: true
    ImageMagick: true
    FVS: true
    xdpyinfo: true
```

TheEvilSkeleton:
Thanks for reporting. Do you know if any change needs to be done on Gamescope's side? I don't have any game to test with, so I can't really do much other than ask questions and have someone test it for me.

SuperSamus:
[...]

PR #2840, opened 2023-04-09
FSR: only set in Gamescope (remove Wine FSR)

SuperSamus:
# Description
FSR settings are moved in Gamescope. Wine FSR isn't set anymore.
Motivation: #2755

TL;DR: Right now, FSR is enabled in two places:
- Wine-GE, that is, only a specific version of Wine. It also already enables it by default, and doesn't even recommend to disable it (#2636).
- Gamescope, which obviously takes effect only if Gamescope is enabled.

First, having two separate FSR "enablers" is messy, as you need logic to avoid enabling both of them at the same time (and Bottles isn't doing that). Considering the options, I opted to simply remove Wine FSR.
Second, the conditions mentioned above aren't communicated at all by having FSR as a standalone setting, which is why I opted to move them into Gamescope's page.

Some comments on what's missing:
- Maybe it should communicate somewhere that Wine-GE already enables FSR, thus if you use it, you shouldn't enable it in Gamescope (still, it wasn't communicated even before).
- Didn't touch .po files yet.

## Type of change

# How Has This Been Tested?
Built Bottles myself, tested if the UI was correct, launched a game.


SuperSamus:
(I'm waiting for comments before doing anything else)

TheEvilSkeleton:
No worries. We've been busy for a few months now, so we're taking our time. I'm in a trip

TheEvilSkeleton:
Can you resolve the conflicts? I did it locally, but I can't figure out how to push it directly to the MR.

TheEvilSkeleton:
Thanks

bottles/frontend/ui/dialog-gamescope.blp
```diff
@@ -179,6 +179,36 @@ template GamescopeDialog : .AdwWindow {
...
+            Adjustment spin_sharpening_strength {
+              step-increment: 4;
+              upper: 20;
+            }
+
+            ;
+          }
+        }
+      }
+
```
TheEvilSkeleton, line 211: In my opinion, it makes more sense to move these above "Miscellaneous", since it's normal to put "Others" sections in the end.

I'm not too fond of the interface. I'm experimenting locally, but I'll see if we could come up with something better.

bottles/frontend/windows/gamescope.py
```diff
@@ -32,9 +32,10 @@ class GamescopeDialog(Adw.Window):
     switch_scaling = Gtk.Template.Child()
     toggle_borderless = Gtk.Template.Child()
     toggle_fullscreen = Gtk.Template.Child()
+    switch_fsr = Gtk.Template.Child()
+    spin_sharpening_strength = Gtk.Template.Child()
     btn_save = Gtk.Template.Child()
     btn_cancel = Gtk.Template.Child()
```
TheEvilSkeleton, line 38: We shouldn't modify 'unrelated' lines for `git blame`.

```suggestion
    btn_cancel = Gtk.Template.Child()

```

TheEvilSkeleton (changes requested):

bottles/backend/models/config.py
```diff
@@ -88,8 +88,7 @@ class BottleParams(DictCompatMixIn):
     gamescope_fullscreen: bool = True
     sync: str = "wine"
     fsr: bool = False
-    fsr_sharpening_strength: int = 2
-    fsr_quality_mode: str = "none"
+    fsr_sharpening_strength: int = 8
```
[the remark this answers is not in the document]
SuperSamus: If you look at other deleted lines you'll see that Wine's FSR and Gamescope's FSR have different ranges (0-5 and 0-20).
Bottles used Wine's range, and multiplied it by 4 when using FSR on Gamescope.
Let's simply use Gamescope's range directly.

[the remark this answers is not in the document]
TheEvilSkeleton: Ah, thanks for the explanation :+1:

bottles/frontend/ui/dialog-gamescope.blp
TheEvilSkeleton: Yeah, I can't think of a better design that fits with the current dialog. In this case, can you move the FSR section above Miscellaneous? I'll just merge it afterwards.

SuperSamus:
One last note. Logs now say:
```
(WARNING) Skipping unexpected config 'fsr_quality_mode' in BottleParams
```

(This config is, in fact, not used anymore.)
Is this log something important to deal with, or not?

koplo199:
Note:
- Since `Wine-GE-Proton8-11`/`GE-Proton8-7` the bug caused by disabling FSR has been [fixed](https://github.com/GloriousEggroll/proton-ge-custom/releases/tag/GE-Proton8-7)
- Since `Wine-GE-Proton8-14`/`GE-Proton8-12` FSR is [disabled by default](https://github.com/GloriousEggroll/proton-ge-custom/releases/tag/GE-Proton8-12)

It adds to the fact that not communicating about `Wine-GE`/`GE-Proton` enabling FSR (between specific, older versions that is) maybe is just fine.

orowith2os:
I think it would be better if we only dealed with gamescope FSR, as Wine FSR can be finnicky, and as you just saw, varies between versions. So we can keep it in the gamescope menu, and remove it otherwise.

SuperSamus:
Rebased and addressed #3065.

orowith2os:
Needs a quick look over from @mirkobrombin, otherwise I'd like to land this.

mirkobrombin (approved):
I checked the changes and looks goof to me. I am not able to test the changes by myself.

Thanks for your effort.

orowith2os (approved):

orowith2os:
Will test later to make sure FSR works; then I'm merging, with some other fixes, for 51.10.

koplo199:
After investigation, I **strongly** disagree with the proposal made here of only setting FSR inside gamescope.
Here is why:

1. Bottles doesn't explicitly disable FSR, it just don't force enable it if not instructed to, proof: [here](https://github.com/bottlesdevs/Bottles/blob/af3a7ccf44cd17d40160b1db66a5c8177b6f4cf5/bottles/backend/wine/executor.py#L80-L82) and [here](https://github.com/bottlesdevs/Bottles/blob/af3a7ccf44cd17d40160b1db66a5c8177b6f4cf5/bottles/backend/wine/executor.py#L80-L82) . So not only the bug doesn't exist in upstream anymore (see my last [comment](https://github.com/bottlesdevs/Bottles/pull/2840#issuecomment-1685212743)), but we weren't even affected by it in the first place. There's no reason to change anything here.
2. We loose the ability to easily set FSR without gamescope.
3. We loose program specific FSR setting. Now, it can only be enabled/disabled globally (and with gamescope only).

Only remains duplicated FSR setting issue, which is a 3 lines fixes at most.

SuperSamus:
1. The goal of this PR wasn't to solve that. Assuming Wine-GE continued to enable FSR by default today, #2636's status wouldn't have changed at all. 
The intention is: by making the FSR setting associated with Gamescope, it is more obvious that Wine-GE still does whatever it wants with FSR.

2. It was also easy to have the setting enabled while it's doing absolutely nothing.
With how messy the `WINE_FULLSCREEN_FSR` situation is, making the user manage it manually with environment variables is more appropriate.
Not everything needs to be an explicit setting. For instance, why is DXVK's frame rate limit not a setting? Because `DXVK_FRAME_RATE` only applies to specific situations, so the user doesn't toggle a setting and get surprised when it does nothing.
IMO, `WINE_FULLSCREEN_FSR` is in a similar situation.

3. Not the biggest deal since it can be toggled with `Super+U`.
I do not disagree that it's a downgrade, but the option *needs* to be in a Gamescope's submenu (making Preference Override do this is out of scope for this PR).

koplo199:
> 1. The goal of this PR wasn't to solve that. Assuming Wine-GE continued to enable FSR by default today, [[Bug]: FSR enabled in Wine-GE when disabled in Bottles #2636](https://github.com/bottlesdevs/Bottles/issues/2636)'s status wouldn't have changed at all.

It got disabled by default at the same time it was fixed, so that issue could be closed safely.

> It was also easy to have the setting enabled while it's doing absolutely nothing.

> With how messy the WINE_FULLSCREEN_FSR situation is making the user manage it manually with environment variables is more appropriate.

It is also easy to call the WINE_FULLSCREEN_FSR situation messy, when in reality this designation only rely on one upstream bug that got fixed in the meantime. Unless there's other bugs I'm unaware of?

> Not everything needs to be an explicit setting. For instance, why is DXVK's frame rate limit not a setting? Because DXVK_FRAME_RATE only applies to specific situations, so the user doesn't toggle a setting and get surprised when it does nothing.

I absolutely agree with this observation. I only disagree with the conclusion: it is fairly common to have a resolution setting/slider in-game. Having FSR kicking in while not on 100% resolution, basically providing better performance with minimal visual loss, is not as niche as `DXVK_FRAME_RATE`. I do not even see a point in setting `DXVK_FRAME_RATE`, maybe except for games with buggy VSYNC implementation.

> Not the biggest deal since it can be toggled with Super+U.

I invested a lot of time researching the topic and this is something I didn't even know. So, first: thanks for the tip :sweat_smile:, and second: I think I take no risk saying the vast majority of users may, just like I was 30 seconds ago, not be aware of the existence of this setting.

> I do not disagree that it's a downgrade, but the option needs to be in a Gamescope's submenu (making Preference Override do this is out of scope for this PR).

I don't understand why you say it _needs_ be in the gamescope's submenu: what is the problem with using the global/program FSR value to decide if FSR is enabled inside gamescope?

Also, something that hasn't been mentioned is that wine FSR is far from exclusive to Wine-GE. For instance the default soda runner also include these [patches](https://github.com/bottlesdevs/build-tools/blob/c0dae7673df897fd38d51c72741ea33e74e5b5fc/runners/vaniglia/wine-tkg-valve.cfg#L22). They are also enabled by default in upstream [wine-tkg](https://github.com/SingleInfiniteLoop/wine-tkg-git/blob/34a641ed97570d93db6410e2a74fa588cf1d2ee6/wine-tkg-git/customization.cfg#L45), meaning it will be available for any runner using tkg patches, assuming of course the said runner does not explicitly set `_use_GE_patches="false"`.

TheEvilSkeleton:
I apologize for the long overdue review of this, as some of us were burned out and needed a break. Is it something you want to continue working on?

TheEvilSkeleton:
Huh, it seems like GitHub doesn't notify me via emails when a push happens. I'm really sorry for being a week late and merging things that ended up conflicting with this MR :/

TheEvilSkeleton:
@SuperSamus @koplo199 I'd like to revive the discussion in https://github.com/bottlesdevs/Bottles/pull/2840#issuecomment-1750397125. I don't really know which decision to go for, but I'm more in favor of accepting this MR, mainly because it cleans up a good chunk of the code. I think it's wise(r) to have the runner decide for the user whether FSR should be enabled or not.

mirkobrombin:
I reviewed all the comments and I am fine with merging this MR. Can you please fix conflicts? Sorry for the delay, I was busy.

SuperSamus:
Whoops, I missed the email.
Here it is.

PR #2886, opened 2023-05-07
Block the use of GTK_THEME in bottles

axtloss:
# Description
Makes Bottles ignore the `GTK_THEME` environment variable.

Fixes any issue that complained about GTK_THEME

## Type of change

# How Has This Been Tested?
- set `GTK_THEME` using flatseal
- restart Bottles
- Bottles still uses the usual libadwaita stylesheet

### DIFF
--- a/bottles/frontend/bottles.py
+++ b/bottles/frontend/bottles.py
@@ -17,6 +17,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import os
+os.environ.pop('GTK_THEME')
 import sys
 import signal
 import gettext


mirkobrombin (approved):

bottles/frontend/bottles.py
```diff
@@ -17,6 +17,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import os
+os.environ.pop('GTK_THEME')
```
TheEvilSkeleton, line 21: Does it have to be here? I would prefer to have it below the imports

axtloss: No it probably does not have to be there, but doing this as early as possible would ensure that gtk is not initialized before removing the environment variable

koplo199:
This is absolutely terrible and anti-freedom. Why even do this?

Kinsteen:
I think maybe being able to enable theming with a env variable like `I_REALLY_WANT_TO_THEME_BOTTLES_AND_I_WONT_COMPLAIN_ABOUT_BUGS` or something? I don't like outright blocking themes as well

axtloss:
This does not block theming, it blocks wrong theming. `GTK_THEME` isn't the proper way to apply themes (and from what I know, doesn't even properly work for libadwaita), if people want to theme bottles they should use `~/.config/gtk-4.0`

There's nothing against user freedom here, we are just making our lives easier with not having to waste our time on people who follow random guides and end up setting `GTK_THEME`.

jannuary:
FWIW theming was already broken (as it always is, but that's a different discussion), this simply provides a failsafe in case someone's using the **explicitly made for debug purposes** GTK_THEME option.

PR #2948, opened 2023-07-02
Use the correct icon in dependency-entry.blp for downloads

### LINKED ISSUE #2949: [Bug]: Bottles is using save icon for downloads instead of download icon
### Describe the bug

using `document-save-symbolic` for downloads is worng everywhere except adwaita
if we use something like `folder-download-symbolic` intead the icon will be correct everywhere.

### To Reproduce

Look at it using any icon theme that isn't adwaita

### Package

Flatpak from Flathub

### Distribution

Non distribution specific

### Debugging Information

```shell
None
```

### Troubleshooting Logs

```shell
None
```

### Additional context

_No response_

CosmicFusion:
# Description
I know bottles is design around Adwaita and Adwaita only.

but save icons are universally agreed to be floppy disks and adwaita is the one in the wrong here to make save a download icon

So bottles should use a download icon instead such as the one in this PR and the icon would be universally correct in all DEs and icon themes

Will not causing any harm

Fixes #2949

## Type of change

# How Has This Been Tested?
- [x] Look at it using the adwaita icon theme.
- [x] Look at it using Papirus and vimix.

### DIFF
--- a/bottles/frontend/ui/bottles.gresource.xml
+++ b/bottles/frontend/ui/bottles.gresource.xml
@@ -61,7 +61,7 @@
         <file alias="applications-system-symbolic.svg">../../../data/icons/symbolic/scalable/actions/applications-system-symbolic.svg</file>
         <file alias="application-x-addon-symbolic.svg">../../../data/icons/symbolic/scalable/actions/application-x-addon-symbolic.svg</file>
         <file alias="computer-symbolic.svg">../../../data/icons/symbolic/scalable/actions/computer-symbolic.svg</file>
-        <file alias="document-save-symbolic.svg">../../../data/icons/symbolic/scalable/actions/document-save-symbolic.svg</file>
+        <file alias="folder-download-symbolic.svg">../../../data/icons/symbolic/scalable/actions/folder-download-symbolic.svg</file>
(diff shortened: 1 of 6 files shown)


orowith2os:
Your wording here makes it seem as if libadwaita and Bottles are hostile to other desktops. Try and think of the reasoning behind things, and the full context - just saying "adwaita is the one in the wrong" helps no one.

Please make an issue with the libadwaita folks over in [adwaita-icon-theme](https://gitlab.gnome.org/GNOME/adwaita-icon-theme/) to see if they'll change document-save to a floppy disk, and if not, their reasoning for why.

CosmicFusion:
> Your wording here makes it seem as if libadwaita and Bottles >are hostile to other desktops.

That's was not my intention

Just seeing "Stop theming our apps" url stamped on the readme

Made me think that validation is only done against adwaita

And I was not sure about the overall stance on this

>Try and think of the reasoning behind things, and the full context - just saying "adwaita is the one in the wrong" helps no one.
>

True...

> Please make an issue with the libadwaita folks over in [adwaita-icon-theme](https://gitlab.gnome.org/GNOME/adwaita-icon-theme/) to see if they'll change document-save to a floppy disk, and if not, their reasoning for why. 

Will do my friend

TheEvilSkeleton:
Drafting this as it's blocked by upstream. When you file the issue, please link here so we can track.

PR #2993, opened 2023-07-25
winecommand: Fix LatencyFleX enablement

Joshua-Ashton:
This is not a driver ICD...


orowith2os:
Have you tested locally (via a flatpak bundle, either from here or a flatpak built locally) that this works?

Joshua-Ashton:
I did not test in the Flatpak but I did on my host. It will work as long as the Flatpak (or it's runtimes) ship LatencyFleX.

Without this commit, LatencyFleX is being added as a driver ICD which is completely wrong, it is a Vulkan layer and the option is not correctly controlling whether the layer is enabled.

orowith2os (approved):
Everything LGTM, and is in line with upstream behavior. Will test momentarily.

orowith2os:
> I did not test in the Flatpak but I did on my host. 

If you could test in Flatpak, that would be greatly appreciated. We don't support native packages *at all*, and if it only works there, this isn't something I think we're going to worry about.

TheEvilSkeleton:
It's our responsibility as maintainers to test it :p

That said, this is a proper fix and in-scope of what it's trying to achieve. If the LatencyFleX flatpak doesn't work then it's a packaging problem on our end; fixing that would make it out of scope of the MR.

Issue #3041, opened 2023-08-19
[Bug]: --version, --help and --help-all shouldn't load GUI/GTK libs


drws:
### Describe the bug

Running Bottles in CLI with `--version`, `--help` or `--help-all` loads GTK. These low-level CLI-only commands should not load any GUI libs (GTK) for most compatibility.

### To Reproduce

1. `bottles -v`
2. observe _standard_ GTK warnings like _Theme parser error_

### Package

Flatpak from Flathub

### Distribution

Arch Linux

### Debugging Information

```shell
no debug info
```

orowith2os:
Most compatibility with what? This definitely shouldn't be happening anyways, at least at first glance - but most compatibility with what?

drws:
Well, CLI. Even though practically all Bottles' functionality is GUI-only and help and version queries don't even do anything concrete, they still provide very basic and informally-standardised _CLI-only_ functionality. There is no reason they shouldn't work in case of wrong/broken/missing GUI dependencies for _most_ CLI compatibility, whatever that is :)

orowith2os:
You shouldn't be encountering issues with dependencies anyways if you're using Bottles as intended, aka not a distro package. 

PRs welcome if you want to fix the loading of GTK libs.

orowith2os:
We seem to be suppressing warnings anyways, not sure why that's not working. 
https://github.com/bottlesdevs/Bottles/blob/9132667ae41a04769b286a09471306ba4deed980/bottles/frontend/cli/cli.py#L30

PR #3045, opened 2023-08-20
Don't load in GTK when using the CLI frontend

orowith2os:
# Description
We shouldn't be loading in GTK when using the CLI frontend, at least not directly.

Fixes https://github.com/bottlesdevs/Bottles/issues/3041

## Type of change

# How Has This Been Tested?
Not yet. Waiting for CI bundles.


orowith2os:
Looks like it's needed, and would some work to port to not use it.

PR #3057, opened 2023-08-24
fix: LatencyFlex

### LINKED ISSUE #2966: [Bug]: LatencyFlex updates are placebo
### Describe the bug

The `LatencyFlex` shared library is included at built time, but is never updated after that:
https://github.com/bottlesdevs/Bottles/blob/9132667ae41a04769b286a09471306ba4deed980/com.usebottles.bottles.yml#L168-L174

Only the [dll's](https://github.com/bottlesdevs/Bottles/blob/9132667ae41a04769b286a09471306ba4deed980/bottles/backend/dlls/latencyflex.py#L25-L26) are updated, which given their sizes I guess are only shims to the native library. Using an unpaired combination of windows dll's and native shared library will probably lead to issues at some point.

I'm still unsure what this implies : does that means different bottles cannot use different versions of `LatencyFlex`? Can `LatencyFlex` even be updated at runtime?

koplo199:
# Description
See #2966
Fixes #2966

## Type of change

# How Has This Been Tested?
- [ ] Test A
- [ ] Test B

### DIFF
--- a/bottles/backend/wine/winecommand.py
+++ b/bottles/backend/wine/winecommand.py
@@ -282,15 +282,17 @@ class WineCommand:
             env.add("__GL_SHADER_DISK_CACHE_PATH", os.path.join(bottle, "cache", "gl_shader"))
             env.add("MESA_SHADER_CACHE_DIR", os.path.join(bottle, "cache", "mesa_shader"))
 
-        # VKD£D environment variables
+        # VKD3D environment variables
         if params.vkd3d and not return_steam_env:
             env.add("VKD3D_SHADER_CACHE_PATH", os.path.join(bottle, "cache", "vkd3d_shader"))
 
         # LatencyFleX environment variables
         if params.latencyflex and not return_steam_env:
             _lf_path = ManagerUtils.get_latencyflex_path(config.LatencyFleX)
-            _lf_icd = os.path.join(_lf_path, "layer/usr/share/vulkan/implicit_layer.d/latencyflex.json")
+            _lf_layer_path = os.path.join(_lf_path, "layer/usr/share/vulkan/implicit_layer.d")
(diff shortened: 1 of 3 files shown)


bottles/backend/wine/winecommand.py
```diff
@@ -282,15 +282,17 @@ def get_env(self, environment: Optional[dict] = None, return_steam_env: bool = F
...
         # LatencyFleX environment variables
         if params.latencyflex and not return_steam_env:
             _lf_path = ManagerUtils.get_latencyflex_path(config.LatencyFleX)
-            _lf_icd = os.path.join(_lf_path, "layer/usr/share/vulkan/implicit_layer.d/latencyflex.json")
+            _lf_layer_path = os.path.join(_lf_path, "layer/usr/share/vulkan/implicit_layer.d")
+            env.concat("VK_ADD_LAYER_PATH", _lf_layer_path)
             env.add("LFX", "1")
+            ld.append(os.path.join(_lf_path, "layer/usr/lib/x86_64-linux-gnu"))
         else:
             env.add("DISABLE_LFX", "1")
```
orowith2os, line 297: So now we pull from the Bottles repos and set it all up here, instead of using what ships with the Bottles flatpak?

koplo199: In short: Yes.

In less short: LatencyFlex has configurable versions just like DXVK and VKD3D-Proton, previously we were in a mixed setup where the dll's libraries were those of the user chosen LatencyFlex version, and the native library from the one shipped with flatpak. Now it correctly uses everything from the version chosen by the user, meaning there's no more risks of an incompatible dll/native libraries couple as before, among other things. It also means we don't need to ship it with flatpak anymore.

orowith2os (approved):

PR #3102, opened 2023-09-27
Minor improvements

koplo199:


bottles/backend/health.py
```diff
@@ -104,38 +93,33 @@ def check_x11(self):
...
+        return "WAYLAND_DISPLAY" in os.environ
 
     def check_xwayland(self):
-        if self.x11 and self.wayland:
-            return True
-        return False
+        return self.x11 and self.wayland
 
     def check_desktop(self):
         return os.environ.get("DESKTOP_SESSION")
```
orowith2os, line 102:
```suggestion
        return os.environ.get("XDG_CURRENT_DESKTOP")
```

`DESKTOP_SESSION` is deprecated, and has been for a while.

bottles/backend/health.py
```diff
@@ -104,38 +93,33 @@ def check_x11(self):
...
+    @staticmethod
+    def check_xdpyinfo():
+        return bool(shutil.which("xdpyinfo"))
+
+    @staticmethod
+    def check_ImageMagick():
+        return bool(shutil.which("identify"))
 
     @staticmethod
     def check_icoextract():
```
orowith2os, line 125: Many other `check_*` methods here are just a `try`, `import`, and a `catch`. Is there no way to make it nicer, and maybe take advantage of Python's dynamic type system here to convert any exceptions to `False`? 

It also looks like mypy might not be getting applied to these files; they don't have a type declaration AIUI.

orowith2os: An alternative method, courtesy of @refi64, because apparently an Exception -> bool conversion ain't happening:

```python
def check_module(module):
  try:
    __import__(module)
  except ModuleNotFoundError:
    return False
  else:
    return True
    ```
    
    And we can get rid of many of these individual `check_*` methods.

Comment (orowith2os) [reply bottles/backend/health.py:125]: And a similar cleanup for all the `schutil` methods.

Comment (koplo199) [reply bottles/backend/health.py:125]: That indeed would be much cleaner, I'll implement the changes.

Comment (orowith2os) [reply bottles/backend/health.py]: We can get rid of each of the individual methods too, not just redirect them.

Comment (koplo199) [reply bottles/backend/health.py]: I have no idea of what you have in mind. I'll close the PR but keep the branch intact so you can still work on it.

Comment (orowith2os) [reply bottles/backend/health.py]: I mean, instead of `check_*`, we do `check_module()` and `check_executable()`, but iterate over the names and whatnot. Thanks for kicking off most of this again anyways; everything else but what I'm talking about can be merged now, if you'd like to reopen and remove the changes.

File: bottles/backend/health.py [hunk 3]
```diff
@@ -104,92 +93,66 @@ def check_x11(self):
...
-        return False
+        return self.x11 and self.wayland
 
     def check_desktop(self):
-        return os.environ.get("DESKTOP_SESSION")
+        return os.environ.get("XDG_CURRENT_DESKTOP")
+
+    @staticmethod
+    def check_executable(executable):
+        return bool(shutil.which(executable))
```
orowith2os, line 106:
```suggestion
    # TODO: additional cleanup: don't have individual `check_*` methods, just one like this,
    # and iterate over them for the results.
    def check_executable(executable):
        return bool(shutil.which(executable))
```

orowith2os (approved):

PR #3112, opened 2023-10-01
fix: MangoHud, gamescope and vkBasalt versions

### LINKED ISSUE #3107: MangoHud 0.7.0 (branch 23.08) doesn't work with Bottles
The new Flatpak MangoHud 0.7.0 version (branch 23.08) doesn't work with Bottles.

koplo199:
It will not work if the versions don't match.

### DIFF
--- a/bottles/backend/globals.py
+++ b/bottles/backend/globals.py
@@ -20,7 +20,7 @@ import shutil
 from functools import lru_cache
 from pathlib import Path
 from typing import Dict
-from bottles.backend.utils import json
+from bottles.backend.utils import yaml, json
 
 @lru_cache
 class Paths:
@@ -79,7 +79,9 @@ vkbasalt_available = Paths.is_vkbasalt_available()
 mangohud_available = shutil.which("mangohud") or False
 obs_vkc_available = shutil.which("obs-vkcapture") or False
 vmtouch_available = shutil.which("vmtouch") or False
-base_version = json.load(open("/app/manifest.json")).get("base-version","").removeprefix("stable-")
+base_version = ""
+if os.path.isfile("/app/manifest.json"):
+    base_version = json.load(open("/app/manifest.json")).get("base-version","").removeprefix("stable-")
 
 # encoding detection correction, following windows defaults
 locale_encodings: Dict[str, str] = {
--- a/bottles/backend/globals.py
+++ b/bottles/backend/globals.py
@@ -20,7 +20,7 @@ import shutil
 from functools import lru_cache
 from pathlib import Path
 from typing import Dict
-
+from bottles.backend.utils import json
 
 @lru_cache
 class Paths:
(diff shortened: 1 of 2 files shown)


orowith2os:
I believe we can actually extract the runtime version from /etc/os-release and list that here, so it's always accurate.

koplo199:
> I believe we can actually extract the runtime version from /etc/os-release and list that here, so it's always accurate.

Unfortunately, only the GNOME runtime version is shown in `/etc/os-release`. There is `/app/manifest.json` from which we could extract the [base-version](https://github.com/bottlesdevs/Bottles/blob/af3a7ccf44cd17d40160b1db66a5c8177b6f4cf5/com.usebottles.bottles.yml#L5), but I don't know if it would be accurate.

bottles/backend/globals.py
```diff
@@ -79,6 +79,9 @@ class TrdyPaths:
 mangohud_available = shutil.which("mangohud") or False
 obs_vkc_available = shutil.which("obs-vkcapture") or False
 vmtouch_available = shutil.which("vmtouch") or False
+base_version = ""
+if os.path.isfile("/app/manifest.json"):
```
orowith2os, line 83: We can safely assume this file will always exist, no need to check for that.

bottles/frontend/views/bottle_preferences.py
```diff
@@ -194,11 +194,15 @@ def __init__(self, details, config, **kwargs):
...
+        _obsvkc_pkg_name = "com.obsproject.Studio.Plugin.OBSVkCapture"
         _flatpak_pkg_name = {
-            "gamescope": "org.freedesktop.Platform.VulkanLayer.gamescope",
-            "vkbasalt": "org.freedesktop.Platform.VulkanLayer.vkBasalt//22.08",
-            "mangohud": "org.freedesktop.Platform.VulkanLayer.MangoHud//22.08",
-            "obsvkc": "com.obsproject.Studio.Plugin.OBSVkCapture"
+            "gamescope": f"{_gamescope_pkg_name}//{base_version}" if base_version else _gamescope_pkg_name,
+            "vkbasalt": f"{_vkbasalt_pkg_name}//{base_version}" if base_version else _vkbasalt_pkg_name,
+            "mangohud": f"{_mangohud_pkg_name}//{base_version}" if base_version else _mangohud_pkg_name,
+            "obsvkc": _obsvkc_pkg_name
```
orowith2os, line 205: Same deal, no need to check `base_version`. Just the `{*_pkg_name}//{base_version}"` bit works. No need to inline the names here in future changes, since that'll just be a bit more pain.

orowith2os:
Well, I fucked up on the /app/manifest.json shenanigans. It looks like our test env doesn't contain it, so it fails. Maybe spoofing it would work.

koplo199:
@orowith2os Yes, I originally did  in 90ee77dbd7a3b8d672fd5083380da57526312d11 what you were suggesting, but had to add that condition in 3717f01940b2980743698e89099c47d8b24cf3ab to fix the CI from failing, for the better in my opinion.

Issue #3106, opened 2023-09-30
[Bug]: If Bottle name contains "<" the name does not display in the main menu


JordanPlayz158:
### Describe the bug

As the title suggests, the bug is that when you include the "<" character in the Bottle name (not sure of any others with this issue but "|", "=", and ">" seem to be fine displaying in bottle name), in the main menu it displays "Bottle name" (but on the bottle page it displays fine), not a big deal but figured I'd document it.

### To Reproduce

Rename a bottle with the "<" symbol included (ex. how I found this, I put the prefix of "DXVK <=2.2 REQUIRED | " in one of my bottle names) and restart (only shows bug after restart)

Example of how the bug looks like
Bottle name: https://cdn.jordanplayz158.xyz/uploads/d5c455cef09c00710f937792c0e4ed8b1409464f.png
Bottle name from Bottles list **after restart**: https://cdn.jordanplayz158.xyz/uploads/55efedd7fb43888f039c94de5a3b49669f06930c.png

### Package

Flatpak from Flathub

### Distribution

Nobara 38 (Fedora 38 fork)

### Debugging Information

```shell
Official Package: true
Version: '51.9'
DE/WM: plasma
Display:
    X.org: true
    X.org (port): :0
    Wayland: true
Graphics:
    vendors:
        amd:
            vendor: amd
            envs:
                DRI_PRIME: '1'
            icd: /usr/lib/x86_64-linux-gnu/GL/vulkan/icd.d/radeon_icd.x86_64.json:/usr/lib/i386-linux-gnu/GL/vulkan/icd.d/radeon_icd.i686.json
    prime:
        integrated: null
        discrete: null
Kernel:
    Type: Linux
    Version: 6.5.5-200.fsync.fc38.x86_64
Disk:
    Total: 12562538496
    Free: 12562382848
RAM:
    MemTotal: 23.4GiB
    MemAvailable: 14.4GiB
Bottles_envs: null
```

### Troubleshooting Logs

```shell
17:45:52 (INFO) Bottles Started! 
17:45:54 (INFO) Performing Bottles checks… 
17:45:54 (INFO) Dxvks found:
         - dxvk-2.3
         - dxvk-2.2
 
17:45:54 (INFO) Vkd3ds found:
         - vkd3d-proton-2.10
 
17:45:54 (INFO) Latencyflexs found:
         - latencyflex-v0.1.1
 
17:45:54 (INFO) Runners found:
         - caffe-7.20
[...]

PR #3140, opened 2023-10-29
Allow special characters in bottle names

cybin:
- disabled Pango markup for bottle names in dialogs
- use bottle names md5 hash for directory creation

# Description
Currently, bottle names in the UI are treated as Pango markup. This causes problems when using special characters like '<'. To fix this, the md5 hash of a bottles name is used to create the directory. This shouldn't be a problem as the bottles name for UI is taken from bottles.yml. So, the user will be able to use those characters now. One downside is that the directory for a bottle is not named like it is shown in the UI. On the other hand, Bottles is a manager for bottles and as a user it shouldn't require me to work manually within the filesystem.

These changes does not affect already existing bottles, as far as I could test it. It just works for newly created ones. Well, this is a kind of brute force attempt and might be tested by others as I'm sure I didn't seen all possible dialogs using a bottles name.

Fixes #(issue)
https://github.com/bottlesdevs/Bottles/issues/3106

## Type of change

# How Has This Been Tested?

- Open 'New Bottle' dialog
- Name the bottle 'This is a nasty < Test with {}[]\/'
- Click create.

### DIFF
--- a/bottles/backend/managers/manager.py
+++ b/bottles/backend/managers/manager.py
@@ -15,6 +15,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
+import hashlib
 import contextlib
 import fnmatch
 import os
@@ -1025,8 +1026,8 @@ class Manager(metaclass=Singleton):
             '''
             config.NVAPI = self.nvapi_available[0]
 
-        # create the bottle path
-        bottle_path = os.path.join(Paths.bottles, config.Name)
+        # create the bottle path; use MD5 hash as path
+        bottle_path = os.path.join(Paths.bottles, hashlib.md5(config.Name.encode()).hexdigest())
 
         if not os.path.exists(bottle_path):
             '''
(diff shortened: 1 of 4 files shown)


xioren:
Is hashing filenames necessary with ```use-markup: false```? Also, minor nitpick...your code comment says hash with MD5 and yet you are hashing with SHA256.

cybin:
> Minor nitpick...your code comment says hash with MD5 and yet you are hashing with SHA256 (which will produce quite long filenames/paths).

argh... initially I used MD5, but it did not pass the quality checks, due to weakness. Yes, I should have removed the comment. I will push an update when I got the time to do that.

The point to use a hash is that using <, ', or the like in paths is a pain. The usual approach is to simply replace them with an underscore or something. But this will lead to '<<' and '>>' be the same '__'. So, as I mentioned before, there shouldn't be need for a 'normal' user to work directly in these directories. And everyone familar to the shell to do so shouldn't have a real problem working with hashes. Well, this is my humble opinion, of course.

orowith2os:
Would disabling the markup here not be enough? Why do we need to also change how directories are created? The less disconnection we have between the filesystem and the UI, the better.

cybin:
> Would disabling the markup here not be enough? Why do we need to also change how directories are created? The less disconnection we have between the filesystem and the UI, the better.

It might be enough to simply disable the markup. There is no need to change the way you create directories, if you don't like.

xioren:
In this case I don't think it is necessary since we already use ```pathvalidate``` to sanitize filenames/paths; otherwise I would support the change.

orowith2os:
I can support this MR without hashing.

PR #3261, opened 2024-01-24
fix(project): specify the corrent gettext domain in appstream file

zefr0x:
# Description
The gettext text domain used in the application and the one which Meson installs to is `bottles`, but the one specified in the appstream file is `com.usebottles.bottles`

https://github.com/bottlesdevs/Bottles/blob/ed765a96599f9596d8010e57b25b3c4ed626a563/bottles/frontend/main.py#L67-L70

https://github.com/bottlesdevs/Bottles/blob/ed765a96599f9596d8010e57b25b3c4ed626a563/po/meson.build#L12

**Also**, should I instead change the gettext domain to `com.usebottles.bottles` in the application and Meson? I think it is best practice, but it might break the Weblate bot or something.

## Type of change

# How Has This Been Tested?
Never tested.


mirkobrombin:
Hi, sorry for the delay. Would be better to move to the RDNN com.usebottles.bottles, since that's the real identifier of the application.

zefr0x:
Sorry about that, github just closed the PR when I synced the branch. I can't reopen it.
Should’ve not used `main` in my fork in the firstplace, this was an old PR.

I already did the work and I will open a new one soon after making sure nothing has to be updated.

PR #3277, opened 2024-02-13
Run_script echo output to console, and capture errors from script

### LINKED ISSUE #3275: [Bug]: Run_script not working
I'm currently experiencing problems getting the Run_script action to work using the information provided by the maintainers guide [here](https://maintainers.usebottles.com/installers/structure/Actions)

Also created a test application based on the information found in issue bottlesdevs/programs#24 
Unfortunately no such luck getting the expected output to show. 

Screenshot_20240210_165515
Screenshot_20240210_165628
Screenshot_20240210_170000

Please can you advise if I'm doing something wrong, it seems that I cannot get this action to work on Kbuntu 23.10, but I haven't tried on any other devices/workstations yet

alextrical:
# Description
run_script steps are now run with output echoed into the terminal, allowing the user to see the status of a script.
Errors raised from the script are now have handling.

Any Raised errors other than 0, will create an error message but continue with the rest of the manifest (non blocking)
However it could be advisable to gracefully terminate the application installation with a GUI message?

Fixes #3275

## Type of change

# How Has This Been Tested?
- [x] Run any application manifest with a "run_script" step. The output will be echoed into the terminal.

### DIFF
--- a/bottles/backend/managers/installer.py
+++ b/bottles/backend/managers/installer.py
@@ -249,13 +249,18 @@ class InstallerManager:
                 return False
 
         logging.info(f"Executing installer script…")
-        subprocess.Popen(
+        result = subprocess.run(
             f"bash -c '{script}'",
             shell=True,
             cwd=ManagerUtils.get_bottle_path(config),
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
-        ).communicate()
+            capture_output=True,
+        )
+        logging.info(f"Output: \n{result.stdout.decode()}\n")
+        if result.returncode != 0:
+            # an error happened!
+            err_msg = "Status: FAIL \nError: %s\nCode: %s" % (result.stderr.decode(), result.returncode)
+            logging.error(err_msg)
+            # raise Exception(err_msg) #Raise blocking error
         logging.info(f"Finished executing installer script.")
 
     @staticmethod


bottles/backend/managers/installer.py
```diff
@@ -249,13 +249,18 @@ def __step_run_script(config: BottleConfig, step: dict):
...
             f"bash -c '{script}'",
             shell=True,
             cwd=ManagerUtils.get_bottle_path(config),
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
-        ).communicate()
+            capture_output=True,
+        )
+        logging.info(f"Output: \n{result.stdout.decode()}\n")
+        if result.returncode != 0:
```
mirkobrombin, line 259: change this to `return result.returncode == 0` this will return `True` if the condition meet and `False` otherwise. If you look at how the `install` function executes the installer' steps, it checks if one of them returns `False` https://github.com/bottlesdevs/Bottles/blob/2bd8b60608661228ea3e804edbd7ab0dc05faa0c/bottles/backend/managers/installer.py#L421 and then return a proper `Result` object according to that. 

Currently, there is no easy way to show a message to the user graphically. It could be possible by making each step returning a tuple (boolean, message) and have the `install` function append that message to the Result object, then updating the installer view in the frontend to use that message if available https://github.com/bottlesdevs/Bottles/blob/2bd8b60608661228ea3e804edbd7ab0dc05faa0c/bottles/frontend/windows/installer.py#L173

alextrical: adding the line `return result.returncode == 0` results in a known failure (exit 1) in the script not even putting a warning in the Console, and just continuing with the rest of the manifest steps. 

Same goes for adding
`return False`
or
```
      return Result( 
         False, data={"message": "Installer is not well configured."} 
     ) 
```
Bottles just continues on with the rest of the manifest regardless of the failure


It looks like there is something missing to handle the return at this point, as it continues regardless of any state returned

mirkobrombin:
It's handled here https://github.com/bottlesdevs/Bottles/blob/2bd8b60608661228ea3e804edbd7ab0dc05faa0c/bottles/frontend/windows/installer.py#L171 it will simply never complete without an ok result. Can' t take a look in a short time.

alextrical:
It seems like I'm missing a couple of tricks today, but as far as i can tell it always completes regardless of the result being set to Null, True or False. I can't get it to fail the install process from a script by setting the "Result"

alextrical:
If you could, that would be great.

I think the next step is working out why the Result doesn't look like it's being checked. As either False or no Result being set don't stop the manifest proceeding to the next step

mirkobrombin:
Closing by no activity.

PR #3303, opened 2024-03-03
metainfo: Add developer

A6GibKm:
developer_name as deprecated in AppStream 1.0, but we cannot remove it as some frontends do not support the newer developer tag.

See https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-developer.


A6GibKm:
You might also consider https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/quality-guidelines#brand-colors.

mirkobrombin:
Thanks for the contribution!

Issue #3423, opened 2024-07-06
[Bug]: taking a backup takes unusually long


red15:
### Describe the bug

When making a bottle backup/export the process will run very slowly as all the file walking/compressing is all in python.

### To Reproduce

1. Go inside a bottle
2. Click secondary menu
3. Take a `Full Backup`
4. Specify the path and begin task.
5. Observe very slow tarfile creation

### Package

Flatpak from Flathub

### Distribution

Arch Linux

### Debugging Information

```shell
Official Package: true
Version: '51.11'
DE/WM: plasmax11
Display:
    X.org: true
    X.org (port): :0
    Wayland: false
Graphics:
    vendors:
        nvidia:
            vendor: nvidia
            envs:
                __NV_PRIME_RENDER_OFFLOAD: '1'
                __GLX_VENDOR_LIBRARY_NAME: nvidia
                __VK_LAYER_NV_optimus: NVIDIA_only
            icd: /usr/lib/x86_64-linux-gnu/GL/vulkan/icd.d/nvidia_icd.json
            nvngx_path: /usr/lib/x86_64-linux-gnu/GL/nvidia-555-58-02/extra/nvidia/wine
    prime:
        integrated: null
        discrete: null
Kernel:
    Type: Linux
    Version: 6.9.7-arch1-1
Disk:
    Total: 16803315712
    Free: 16803135488
RAM:
    MemTotal: 31.3GiB
    MemAvailable: 18.0GiB
Bottles_envs: null
```

### Additional context

I'm not very familiar with this project on the Packaging side (I only _use_ Flatpak) but would it perhaps be a better idea to spawn a subprocess using either the `tar` command directly.

This would bring some good features:
* backups would be asynchronous (ie Bottle executable is not the one doing the IO/CPU)
* The task can become a `cancellable` task (and maybe even get some progress on the created file size in the subtitle?)
* the subprocess could in theory survive the parent process being terminated ensuring a complete backup
  * but this might not be desirable as it would leave a running process that the user cannot "manage" without having to resort to system task management tools (ps/top/kill/...)
  * also probably doesn't work in a flatpak environment

red15:
Doing some more tests it seems most of the speed is actually lost in the gzip part due to my bottle being 250Gb of mostly already compressed data.

PR #3424, opened 2024-07-06
Added extra method of making backups faster and report progress.

red15:
This depends on external binary `tar` being available but is being tested for availability and falls back gracefully.

# Description
To make backups faster and use external tar for simpler / faster backup creation.

Fixes #(issue)
#3423 

## Type of change

# How Has This Been Tested?
Completely untested, was hoping there is CI available or someone can run the test easily?

### DIFF
--- a/bottles/backend/managers/backup.py
+++ b/bottles/backend/managers/backup.py
@@ -54,9 +54,11 @@ class BackupManager:
         try:
             import subprocess
             with subprocess.Popen([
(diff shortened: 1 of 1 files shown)


red15:
Also just realized that the resulting tarball might not be a valid backup, but for this it should be a simple matter of tweaking some of the tar options (such as `f"--directory={source_path}"`)

PR #3572, opened 2024-11-28
Add Gamescope toggle as an override for individual programs

### LINKED ISSUE #3377: [Request]: Make gamescope an executable/shortcut override

Bottles is the first wine experience I have ever had that allowed me to set up a game in a fully modded state without driving me insane and everything has been perfect except for one caveat.

If gamescope is enabled for the bottle, then it also forces gamescope on all executables, which near always breaks configuration applications, even with virtual desktop enabled in their overrides.

Sonic Adventure DX and SA Mod Manager is the success that I had with my first experience with bottles, and I only struggled with no longer being able to use the game's native configuration app (for controllers) and SA Mod Manager at some point, then rummaging through the options to find out that it was gamescope.

Add Gamescope as an override for executables/shortcuts. A configuration app does not need to be in a gaming/fullscreen optimized compositor. This way, in the case of games like Sonic Adventure DX, I can leave gamescope enabled for sonic.exe itself but disable it for the native config app and mod manager.

### Other solutions?

For now I just leave gamescope disabled, or I toggle it back on once configuration is done. It works but its tedious.

_No response_

EmoonX:
# Description

Allows Gamescope to be toggled on a program-by-program basis, instead of just bottle-wide. This means that you can either leave the global toggle off and enable it for individual programs (default settings will be used) or leave it global-on and disable on specific ones.

image

All in all, avoids the hassle of having to manage two near-identical instances of the same bottle just to deal with gamescope-ness.

Closes #3377 (well-requested feature).

## Type of change

# How Has This Been Tested?

Try toying with different values for Gamescope (global setting and individual override):
- Override not set (`gamescope: null` in `bottle.yml`) should just be ignored
- Override toggled on (`gamescope: true`) should _always_ start said game with Gamescope
- Override toggled off (`gamescope: false`) should _never_ start said game with Gamescope

### DIFF
--- a/bottles/backend/wine/winecommand.py
+++ b/bottles/backend/wine/winecommand.py
@@ -108,8 +108,8 @@ class WineCommand:
         self.cwd = self._get_cwd(cwd)
         self.runner, self.runner_runtime = self._get_runner_info()
         self.gamescope_activated = (
-            self.config.Parameters.gamescope
-            or environment.get("GAMESCOPE", "0") == "1"
+            environment["GAMESCOPE"] == "1" if "GAMESCOPE" in environment
+            else self.config.Parameters.gamescope
         )
         self.command = self.get_cmd(
             command, pre_script, post_script, environment=_environment
--- a/bottles/backend/wine/winecommand.py
+++ b/bottles/backend/wine/winecommand.py
@@ -107,6 +107,10 @@ class WineCommand:
         self.arguments = arguments
         self.cwd = self._get_cwd(cwd)
         self.runner, self.runner_runtime = self._get_runner_info()
+        self.gamescope_activated = (
+            self.config.Parameters.gamescope
+            or environment.get("GAMESCOPE", "0") == "1"
+        )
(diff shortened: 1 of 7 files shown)


bottles/backend/wine/winecommand.py
```diff
@@ -107,6 +107,10 @@ def __init__(
         self.arguments = arguments
         self.cwd = self._get_cwd(cwd)
         self.runner, self.runner_runtime = self._get_runner_info()
+        self.gamescope_activated = (
+            environment["GAMESCOPE"] == "1" if "GAMESCOPE" in environment
+            else self.config.Parameters.gamescope
+        )
```
EmoonX, line 113: I found this to be the simplest way to check (though not sure if the most orthodox one).

mirkobrombin (approved):
LGTM

bottles/backend/wine/winecommand.py
mirkobrombin: Looks good

PR #3592, opened 2024-12-11
Integrate pre-commit

TheEvilSkeleton:
# Description
This uses [pre-commit](https://github.com/pre-commit/pre-commit) along with other utilities like [MyPy](https://github.com/pre-commit/mirrors-mypy), [ruff](https://github.com/astral-sh/ruff), and [autoflake](https://github.com/PyCQA/autoflake)

@mirkobrombin is this something we want to do?


mirkobrombin:
Why not?

PR #3597, opened 2024-12-13
Revert "feat: add donate button"

TheEvilSkeleton:
This reverts commit b90c9b882fde73a4f41b4a5f15cad6961e20a97d.

# Description
There's already a donate button in the about dialog, which is the appropriate and non-distracting place. It's best not to distract users with pretty colors and whatnot

### DIFF
--- a/bottles/frontend/ui/style-dark.css
+++ b/bottles/frontend/ui/style-dark.css
@@ -2,23 +2,15 @@
   background-color: #d8d8d8;
   color: #000000;
 }
-
 .grade-Gold {
   background-color: #8d7637;
   color: #f2edde;
 }
-
 .grade-Silver {
   background-color: #6b6b6b;
   color: #e8e8e8;
 }
-
 .grade-Bronze {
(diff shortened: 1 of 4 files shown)


EmoonX:
What about swapping pinky (🩷) for whitey (🤍)?

TheEvilSkeleton:
We already have a Donate button in the about dialog:
image

I don't think it's really necessary to have it there

EmoonX:
Haven't donations become more urgent matters as of late, though?
[https://usebottles.com/posts/2024-12-02-future-of-bottles/](https://usebottles.com/posts/2024-12-02-future-of-bottles/)

So a dedicated icon works as a way to draw further attention to that.

TheEvilSkeleton:
Having it in the primary menu should cover that case without being too distracting. As someone with a high probability of having ADHD, having those buttons displayed right there is pretty distracting. I think the primary menu approach would work better because it's less distracting while being still there

PR #3610, opened 2024-12-16
Use `string.ascii_uppercase` instead of exhaustive list for drive letters

EmoonX:
# Description
Saves... 27 lines of code or so?

# How Has This Been Tested?
image

### DIFF
--- a/bottles/frontend/windows/drives.py
+++ b/bottles/frontend/windows/drives.py
@@ -15,7 +15,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
-from string import ascii_uppercase
+import string
 
 from gi.repository import Gtk, GLib, Adw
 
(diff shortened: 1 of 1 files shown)


Vingian:
I noticed there's no C in the original code... probably because it's the main drive.
So... maybe something like `__alphabet = list(ascii_uppercase.replace('C', ''))` ?

EmoonX:
@Vingian I actually completely missed that, but thinking about it now, bear with me: drive C is defined somewhere else as a special case (see [here](https://github.com/bottlesdevs/wine/blob/a42a3169b2763e1daa4ea9311da3372e319b56f7/programs/winecfg/driveui.c#L339) and [here](https://github.com/bottlesdevs/Bottles/blob/53dc73e6885c261b17d56a46f4dda767ea3bf898/bottles/frontend/windows/drives.py#L49)). It always comes by default, and you can't remove it from the drive list nor change its path. Either including `C` or not in the alphabet list makes no difference.

So perhaps it's for the best to simply not hardcode an additional exception. We merely define here all drive letters (from A to Z), and then the program takes care of the rest.

EmoonX:
I noticed something. Will get back to it shortly.

PR #3616, opened 2024-12-19
Fix missing combo letter when removing a drive

EmoonX:
# Description

A replacement to #3610 ("digging gold and finding diamonds" moment).

Fix an issue where removing a drive wouldn't have its letter appearing in the combo list (which means you'd need to do the whole process of restarting Bottles to be able to add it again).

Also refactored and simplified some stuff; e.g. `__alphabet` isn't a volatile list anymore for eager lookup, but just a fixed `"ABCDEFGHIJKLMNOPQRSTUVWXYZ"` string.

## Type of change

# How Has This Been Tested?

Just try adding/removing a bunch of drives. Combo letters (unused ones) and drive letters (used ones) should form a partition of the alphabet (`A` to `Z`), i.e. each and every letter should belong to one and only one of the groups at any time.


mirkobrombin (approved):
LGTM

jntesteves (approved):
LGTM

PR #3625, opened 2024-12-22
Update bn.po

olumolu:
# Description
I will update the bn.po file slowly as i do it i will push a pr consequently and complete this...


olumolu:
> # Description
> 
> I will update the bn.po file slowly as i do it i will push a pr consequently and complete this...

PR #3651, opened 2025-01-01
Auto-mount disc images (CD/DVD) on program launch

### LINKED ISSUE #959: [Request] A way to mount / unmount CD images when launching a program
**Is your feature request related to a problem? Please describe.**
I am looking for a way to trigger a CD image mount before launching a program.  In Lutris, i used "Pre-launch script" and "post-exist" script with simple bash script to trigger cdemu mounts / unmount. But i have not found a way in bottles to do this kind of thing.   

**Describe the solution you'd like**
The best solution would be a cdemu (best cd drive emulator in Linux) integration. Something like a "Mount image" option in Program options.  

**Describe alternatives you've considered**
The pre-launch / post-exit launch bash script like in Lutris is a cheaper solution but it works. It's just kind of a pain to have to create sh script file for each installed game only for that.  

**Additional context**
Just discovered Bottles yesterday and i found it very promising so i'm planning to migrate from Lutris to Bottles, mainly because i use lutris only as a Wine frontend. I'm a kind of old school gamer, i don't use any game platforms (gog, steam etc.) but my old physical games converted to CD ISO images. 

My mount script is as simple as:

    #!/bin/bash
    cdemu load 0 '<iso-filepath>'
    sleep 3
    udisksctl mount -b /dev/sr0
    sleep 3

It could be better i suppose, but it does not require root.

### LINKED ISSUE #2960: [Request]: mount an iso file within the bottle

i would like to install a program, which needs an iso file to be mounted in 'windows' during the installation but it is not possible using bottles i believe. Is there any work around?

a button with 'mount iso-file'

### Other solutions?

_No response_

_No response_

EmoonX:
# Description
This is just a draft (for now).

If successful, should fix #959 and fix #2960.

## Type of change

# How Has This Been Tested?
[REDACTED]

### DIFF
--- a/bottles/backend/managers/manager.py
+++ b/bottles/backend/managers/manager.py
@@ -725,6 +725,7 @@ class Manager(metaclass=Singleton):
                     "icon": "com.usebottles.bottles-program",
                     "pre_script": _program.get("pre_script"),
                     "post_script": _program.get("post_script"),
+                    "disc_image": _program.get("disc_image"),
                     "dxvk": _program.get("dxvk"),
                     "vkd3d": _program.get("vkd3d"),
                     "dxvk_nvapi": _program.get("dxvk_nvapi"),
--- a/bottles/backend/wine/drives.py
+++ b/bottles/backend/wine/drives.py
@@ -33,7 +33,7 @@ class Drives:
         return None
 
     def set_drive_path(self, letter: str, path: str):
-        """Change a drives path in the bottle"""
+        """Change a drive's path in the bottle"""
         letter = f"{letter}:".lower()
         drive_sym_path = os.path.join(self.dosdevices_path, letter)
         if not os.path.exists(self.dosdevices_path):
--- a/bottles/backend/wine/executor.py
+++ b/bottles/backend/wine/executor.py
@@ -35,6 +35,7 @@ class WineExecutor:
         move_upd_fn: callable = None,
         pre_script: Optional[str] = None,
         post_script: Optional[str] = None,
+        disc_image: Optional[str] = None,
         monitoring: Optional[list] = None,
         program_dxvk: Optional[bool] = None,
         program_vkd3d: Optional[bool] = None,
@@ -64,6 +65,7 @@ class WineExecutor:
         self.environment = environment
         self.pre_script = pre_script
         self.post_script = post_script
+        self.disc_image = disc_image
         self.monitoring = monitoring
         self.use_gamescope = program_gamescope
         self.use_virt_desktop = program_virt_desktop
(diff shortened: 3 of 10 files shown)


orowith2os:
The use of flatpak-spawn makes me want to reject this. There should be something else we can do to support disc images... @TheEvilSkeleton?

orowith2os:
Maybe, does Wine allow mounting in specific directories as devices within its filesystem....? Then we could do this without flatpak-spawn.

EmoonX:
You don't have to "reject" anything as this is still just a barebones draft. `flatpak-spawn` is more like a placeholder, I was also using it in another PR until figuring out a library-driven solution is way better (no need to blundle/invoke external commands). So maybe [this one here](https://github.com/libfuse/libfuse) would be the way to go.

Regardless, I didn't actually have luck with `fuseiso` at all. Games overall don't recognize CDs mounted with it... as opposed to what happens when using `gnome-disk-image-mounter` or `cdemu`.

orowith2os:
FUSE doesn't work under Flatpak, I don't think. Like I said, the best option is probably figuring out how to mount directories as drives in Wine, and asking the user to extract their disc images.

SoongVilda:
If you're planning to *still* use Flatpak, here's a not-so-straightforward solution for this issue:

**a)** Use system-available tools for decompressing ISO, IMG, or whatever format requires `sudo`. Can Bottles run under `sudo`? Uh, big no!
**b)** Use FUSE, which can decompress ISO, IMG, or whatever format without `sudo`, but FUSE can't work under Flatpak.
**c)** Decompress .ISO, .IMG, or whatever format it is... and copy all content into the `drive_c` directory within the Wine prefix.

Note: If you use a native build of Bottles like the *unofficial AUR package*, you have a much easier and more convenient way:
**a)** Users can mount ISO files and they will work. **For example:** `sudo mount -t iso9660 -o loop /path/to/file.iso /mnt/iso/`. This is also the best approach, especially if you have an older machine with a DVD/CD device, as many older games use physical media. (Attempt to reslove that under flatpak? Huh... good luck)
**b)** FUSE works without sudo and works system-widely.

In summary, if you need this functionality in the Flatpak version, users must take an extra step: manually decompressing ISO or IMG files themselves. Since I began using the Flatpak version after 2 years with the unofficial AUR package, I've always had to decompress ISO and IMG files manually.

TheEvilSkeleton:
> The use of flatpak-spawn makes me want to reject this. There should be something else we can do to support disc images... @TheEvilSkeleton?

As @EmoonX stated, there's no need to reject anything while it's still in draft. I presume they're experimenting, brainstorming, and learning as they work on this feature, so they're starting with a broad and easy-to-develop implementation, and with the experience gained from it, it allows them to narrow it down.

orowith2os:
We're not going to merge an implementation that relies on flatpak-spawn
either way, but extracting the images would be fine. We just need a UI for
it.

mirkobrombin:
Closing due to no activity.

Issue #3036, opened 2023-08-18
[Request]: Allow changing 'Bottle Directory' for preset types


drws:
### Tell us the problem or your need

Custom settings in _Create New Bottle_ are currently greyed-out when a preset _Application_ or _Gaming_ environment type is selected. However, there is no reason these preset bottles can't be in a custom path.

### Describe the solution you'd like

The `Bottle Directory` setting is never greyed-out and stays available also when _Application_ or _Gaming_ environment is selected.

PR #3653, opened 2025-01-02
Rework New Bottle dialog

TheEvilSkeleton:
Closes https://github.com/bottlesdevs/Bottles/issues/3036

image


EmoonX:
As discussed before, we do agree that list option subtitles are better without periods, but you're missing the `.`s in the settings below, which should be there for consistency with the rest of the application.

image

TheEvilSkeleton:
I think it's best to update the other labels later

TheEvilSkeleton:
Yes, because it follows closely to the human interface guidelines. It only makes sense to use periods when they're multiple sentences

EmoonX:
That doesn't appear to be the standard out there. I checked VS Code, Discord and Firefox as examples... they _all_ use trailing periods for settings' subtitles. Therefore, it's at least debatable.

TheEvilSkeleton:
I was talking about the GNOME HIG: https://developer.gnome.org/hig/guidelines/writing-style.html

EmoonX:
Bottles is (currently) a GTK application, so that's a fair point.

PR #3679, opened 2025-01-07
envvars-dialog: Allow to paste var=value and fix validation

jntesteves:
* Rework the envvars dialog to allow copy/pasting complete variable assignments instead of having to input first the name and later the value.
* Fix the input validation which had some inverted logic and make it behave as expected.
* Make the panel leaner by removing unnecessary headers and periods at the end of labels.

Screenshot From 2025-01-06 20-44-08


bottles/frontend/ui/dialog-env-vars.blp
```diff
@@ -1,40 +1,31 @@
...
-  default-width: 500;
-  default-height: 500;
-
-  ShortcutController {
-    Shortcut {
-      trigger: "Escape";
-      action: "action(window.close)";
-    }
-  }
+template $EnvVarsDialog: Adw.Dialog {
```
TheEvilSkeleton, line 4: I think it'd be best to change the name of the dialog to `BottlesEnvironmentVariablesDialog`. It's definitely long, but I think avoid abbreviations helps the readability. Prefixing with `Bottles` is also important because libadwaita and GTK follow the same convention.

bottles/frontend/windows/envvars.py
```diff
@@ -36,53 +40,83 @@ def __init__(self, parent, env, **kwargs):
...
-        self.set_title(self.env[0])
-        self.set_text(self.env[1])
+        self.set_text("=".join(self.env))
 
         # connect signals
+        self.connect("changed", self.__validate)
         self.connect("apply", self.__save)
         self.btn_remove.connect("clicked", self.__remove)
 
+        row_wg = self.get_child().get_first_child().get_next_sibling()
```
TheEvilSkeleton, line 50: I think `wg` shouldn't be abbreviated.

bottles/frontend/windows/envvars.py
```diff
@@ -36,53 +40,83 @@ def __init__(self, parent, env, **kwargs):
...
+        child_wg = row_wg.get_first_child() if row_wg else None
```
TheEvilSkeleton, line 51: `get_first_child` returns NULL (None) if a child widget doesn't exist.

bottles/frontend/ui/env-var-entry.blp
```diff
@@ -2,12 +2,12 @@ using Gtk 4.0;
 using Adw 1;
 
 template $EnvVarEntry: Adw.EntryRow {
```
TheEvilSkeleton, line 4: This should be renamed to `BottlesEnvironmentVariableEntry`.

TheEvilSkeleton (changes requested):

bottles/frontend/windows/envvars.py
jntesteves: I added a try/except block around this code to catch any errors that could be raised here, since we can safely ignore them all.

bottles/frontend/ui/dialog-env-vars.blp
jntesteves: Those libs use their namespaces because they are shared across many projects, but I don't think we need to do that in client code, it's all obviously Bottles, the prefix is unnecessary IMO. I used just `EnvironmentVariablesDialog` like the rest of our classes, I think that's better.

TheEvilSkeleton: Fair enough. I personally prefer the namespace for consistency reasons, but I guess I will have to change it back to how it was everywhere else

bottles/frontend/ui/env-var-entry.blp
jntesteves: Done

jntesteves: Do you think we should also rename the files to longer names? Currently, they are `dialog-env-vars.blp` and `env-var-entry.blp`.

TheEvilSkeleton: Yeah... I don't really know how to approach it. I do agree that there should be a standard that makes more sense. I only follow that convention on things I recently rework on

TheEvilSkeleton (changes requested):
Some things that should be fixed:

- Inserting a character and removing it will keep showing the check button:
  image
  Instead, the check button shouldn't appear.
- The moment you type something, the row turns red, and the check appears:
  image
  Instead, the row should keep the neutral color without displaying the check. The error should only appear when the user explicitly adds the equal sign (`=`) and removes it.
- If you remove the last defined environment variable, the "No environment variables defined" label doesn't appear, when it should.
- It's pretty easy to accidentally close the dialog without saving your changes, so there could be an [AlertDialog](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.AlertDialog.html) to confirm the user intends to discard.
- I think the dialog is too big for what it is, perhaps you could adjust the height to `content-height: 550;` and `content-width: 500;`?

TheEvilSkeleton:
Good point on every disagreement. I'll take a look at the code

bottles/frontend/windows/envvars.py
```diff
@@ -36,53 +40,90 @@ def __init__(self, parent, env, **kwargs):
...
+                widget = widget.get_next_sibling()
+
+            if isinstance(widget, Gtk.Text):
+                widget.set_valign(Gtk.Align.CENTER)
+            else:
+                raise RuntimeError("Could not find widget Gtk.Text")
+        except Exception as e:
+            logging.error(
+                f"{type(e)}: {e}\nEnvironmentVariableEntryRow could not find text widget. Did AdwEntryRow change it's widget tree?"
+            )
```
TheEvilSkeleton, line 65: I think leaving a comment explaining why this hack is necessary would help future readers

TheEvilSkeleton (changes requested):

bottles/frontend/windows/envvars.py
```diff
@@ -36,53 +40,90 @@ def __init__(self, parent, env, **kwargs):
...
         """
         Remove the env var from the bottle configuration and
         destroy the widget
         """
+        self.__remove_config()
+        self.parent.remove_entry(self)
+
+    def __remove_config(self, *_args):
+        """
+        Remove the env var from the bottle configuration
```
TheEvilSkeleton, line 97: env var → environment variable

TheEvilSkeleton (changes requested):

bottles/frontend/utils/gtk.py
```diff
@@ -15,31 +15,40 @@
...
-        if (
-            re.search("[@!#$%^&*()<>?/|}{~:.;,'\"]", text)
-            or len(text) == 0
-            or text.isspace()
-        ):
+        var_assignment = entry.get_text()
+        var_name = ShUtils.split_assignment(var_assignment)[0]
+        if var_name and not ShUtils.is_name(var_name):
+            entry.set_show_apply_button(False)
+            entry.set_show_apply_button(True)
```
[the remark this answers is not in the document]
jntesteves: It is intentional, but it's a fair question, the code should inform the reader when such weird choices are made. I'll put a comment on it so no one tries to "fix" it in the future.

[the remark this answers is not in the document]
jntesteves: Moved to method with docstring.

bottles/frontend/windows/envvars.py
jntesteves: Moved to method with docstring.

bottles/frontend/windows/envvars.py
jntesteves: Done, for all docstrings.

TheEvilSkeleton (approved):
Everything looks good to me. The only thing that needs to be done is updating the `POTFILES` file. After that, you're free to merge.

jntesteves:
> Everything looks good to me. The only thing that needs to be done is updating the `POTFILES` file. After that, you're free to merge.

Is this what you mean? I've regenerated the translation files. It seems that hasn't been done in a while, it made hundreds of changes coming from many changed files.

TheEvilSkeleton:
My apologies, I thought you changed the file names but my eyes deceived me. Feel free to revert it :sweat_smile:

PR #3683, opened 2025-01-08
Added support for Ghostty

lucas-yotsui:
# Description
Added support for running applications using Ghostty.

_Note: I don't know why the other terminals require the 'sh -c' or
'bash -c', but Ghostty calls that already when passing the option '-e'.
It worked fine in my pc without that, but it would be nice if someone
else tested it just to make sure._

## Type of change

# How Has This Been Tested?
- [ ] Built and installed
- [ ] Removed EasyTerm (installed by the flatpak builder) and any other terminal emulators from my machine
- [ ] Ran executable checking the "Run in Terminal" option

Worked fine here, but as I mentioned further testing would be nice.

### DIFF
--- a/bottles/backend/utils/terminal.py
+++ b/bottles/backend/utils/terminal.py
@@ -43,6 +43,7 @@ class TerminalUtils:
         # Third party
         ["foot", "%s"],
         ["kitty", "%s"],
+        ["ghostty", "-e %s"]
         ["tilix", "-- %s"],
         # Desktop environments
         ["xfce4-terminal", "-e %s"],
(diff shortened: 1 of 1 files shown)


orowith2os:
@TheEvilSkeleton looks like something to patch up with the recent patches to focus on the Flatpak.

TheEvilSkeleton:
I didn't even know we had that code around. I wouldn't mind merging this because it's the backend side of things, but I'm not entirely sure if the code is even correct?

TheEvilSkeleton:
I think the bigger issue is that we're hard coding all these terminals, when we should instead be letting the user set it themselves

lucas-yotsui:
> I think the bigger issue is that we're hard coding all these terminals, when we should instead be letting the user set it themselves

That was actually my first thought upon seeing this code. 

Personally, I think it would be nicer to have a setting like "terminal command" in which I setup the command that works for my emulator, whatever it may be.

lucas-yotsui:
I was cleaning up my Github repos recently and I deleted this fork. Honestly, I consider my additions a really terrible workaround to avoid fixing the real issue pointed by @TheEvilSkeleton. 

I would try to fix this myself, but I'm really neither a huge Python fan nor a Flatpak user (the sheer memory of trying to compile this project gives me chills).

I really like the project though. Honestly hope you guys figure this out, best of luck to you all!

PR #3726, opened 2025-01-25
Remove NVIDIA Flatpak Workaround

404NetworkError:
# Description
Removes the workaround implemented specifically for the NVIDIA Flatpak drivers (5627b5b).

Originally, the issue was brought up in flathub/org.freedesktop.Platform.GL.nvidia#112. flathub/org.freedesktop.Platform.GL.nvidia#331 solves the problem on the driver side.

## Type of change

# How Has This Been Tested?
In the debug information, under `Graphics -> vendors -> nvidia -> icd`, there will be a list of Vulkan `.json` files.
- [X] Using the latest version of `Bottles` from Flathub shows only the `x86_64` file
- [X] Using my branch's build of `Bottles` **still** shows only the `x86_64` file

### DIFF
--- a/bottles/backend/utils/vulkan.py
+++ b/bottles/backend/utils/vulkan.py
@@ -19,7 +19,6 @@ import os
 from glob import glob
 import shutil
 import subprocess
-import filecmp
 
 
 class VulkanUtils:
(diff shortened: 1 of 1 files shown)


TheEvilSkeleton:
I don't use an NVIDIA GPU, so I can't confirm nor deny whether this works.

orowith2os:
Thanks!

PR #3845, opened 2025-04-12
flatpak: Split dependencies

TheEvilSkeleton:


orowith2os (approved):
LGTM, next step is to add this repository as a submodule to the Flathub repo and set up a data checker there that updates the submodule and modifies the Devel manifest appropriately for Flathub.

PR #3966, opened 2025-06-28
Detect Proton from `/usr/share/steam/compatibilitytools.d/` and `$XDG_DATA_HOME/Steam/compatibilitytools.d/`

HimDek:
# Description
Having proton-ge and other custom versions in `/usr/share/steam/compatibilitytools.d/` makes it available in steam. This patch should also make them usable in bottles without having to redundantly store the same version inside bottles.

* Checks subfolders in `/usr/share/steam/compatibilitytools.d/` and `$XDG_DATA_HOME/Steam/compatibilitytools.d/` and add only proton folders to available_runners
* Overrid get_runner_path() method to handle the additional folders

Fixes #(issue)

## Type of change

# How Has This Been Tested?
- [x] Test with non-bottles provided proton
- [x] Install proton outside of bottles
    - [x] Install proton-ge or any version of proton at `/usr/share/steam/compatibilitytools.d/`.
    or
    - [ ] Use steam to install official proton at `$XDG_DATA_HOME/Steam/compatibilitytools.d/`. (Not tested because as far as I know flatpak build will not be able to access data of flatpak steam. However, if installed natively, it should be able to detect proton from native steam)
 - [x] Ensure `flatpak override <Application ID of Bottles> --filesystem=host`.
 - [x] Proton installed during the first steps should be visible and usable inside bottles.

### DIFF
--- a/bottles/backend/managers/manager.py
+++ b/bottles/backend/managers/manager.py
@@ -359,12 +359,26 @@ class Manager(metaclass=Singleton):
         
         if os.path.isdir(Paths.steam_runners):
             steam_runners = glob(f"{Paths.steam_runners}/*/")
+            for runner in steam_runners.copy():
+                if not SteamUtils.is_proton(runner):
+                    logging.warning(
+                        f"Found non-Proton runner in Steam Runners: {runner}. "
+                        "This is not supported and will be ignored."
+                    )
+                    steam_runners.remove(runner)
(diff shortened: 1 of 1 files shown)


mirkobrombin:
Hi, can u re-make this PR on top of main, please? Sorry for the trouble.

PR #3984, opened 2025-07-15
Donation dialog

mirkobrombin:
# Description
This PR adds an AdwMessageDialog prompting for donations. It reads the oldest entry from the Journal Manager and shows the dialog if it’s older than 7 days. Users can choose to hide this message permanently, that preference is stored in data.yml.

image

## Type of change

### DIFF
--- a/bottles/backend/managers/data.py
+++ b/bottles/backend/managers/data.py
@@ -17,7 +17,7 @@
 
 import contextlib
 import os
-from datetime import datetime, timedelta
+
 
 from bottles.backend.globals import Paths
 from bottles.backend.logger import Logger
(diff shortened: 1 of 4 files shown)


francescomasala:
For me is fine!

francescomasala (approved):

pietrodicaprio (approved):
❤️🧡💙

jntesteves:
The Journal has automatic rotation, I think the new dialog will rarely be displayed. I didn't check the exact rules for rotation but from just looking at the log files I think a user who uses Bottles every day will never see the donation dialog.

**EDIT 2025-08-22**: Over a month later, this week I finally saw the dialog for the first time, because I haven't used Bottles for a few days. There's a bit of randomness, but I kinda like that! So heavy users might wait longer, but given enough time, everybody should eventually see the dialog.

PR #4108, opened 2025-10-14
Add playtime tracking backend

evertonstz:
Backend work to implement this request #1273

<s>Used my previous PR as base #4102, I can try to separate them if needed</s>

# Playtime Tracking – Backend PR

## Summary

This PR introduces robust, low-overhead playtime tracking to Bottles’ backend. It records per-program sessions, maintains a fast aggregate table for the UI, and is resilient to crashes and concurrent sessions. The solution is event-driven (signals), easy to toggle via GSettings, and thoroughly tested.

## Motivation

- Provide reliable per-program playtime, last played timestamps, and session history for present and future UI features
- Ensure minimal write amplification and resilience to abrupt exits
- Keep the backend decoupled from execution flow through signals instead of tight coupling

## Design Overview

- Core tracker: `ProcessSessionTracker` (`bottles/backend/managers/playtime.py`)
  - Single SQLite database: `$XDG_DATA_HOME/bottles/process_metrics.sqlite`
  - Tables: `sessions` (history) and `playtime_totals` (materialized aggregate)
  - Heartbeats update `last_seen` periodically
  - Recovery on startup marks unfinished sessions as forced using `last_seen`
  - Totals are kept in sync on session finalization and recovery

- Event-driven lifecycle:
  - Signals added in `bottles/backend/state.py`:
    - `Signals.ProgramStarted`
    - `Signals.ProgramFinished`
  - Payloads typed in `bottles/backend/models/process.py`:
    - `ProcessStartedPayload`
    - `ProcessFinishedPayload` (`status: Literal["success", "unknown"]`)
  - `WineExecutor` emits these signals instead of calling the tracker directly
  - `Manager` subscribes to signals and calls the tracker

- Public backend API consistency
  - New `Manager.playtime_start(...) -> Result[int]` and `Manager.playtime_finish(...) -> Result[None]`
  - Signals and public APIs consistently use `Result[T]`

- Concurrency & durability
  - One tracker instance per process (owned by `Manager`)
  - Single tracker instance is able to reliably monitor multiple apps initiated via bottles GUI (no matter if in the same or different bottles), tho heartbeat is synced between all apps 
  - Single SQLite connection with WAL and busy timeout
  - A global `_lock` guards all DB operations and `_tracked` updates
  - Atomic finalization: session UPDATE and totals UPSERT happen in one transaction
  - Graceful shutdown runs a final WAL checkpoint to avoid leftover `-wal`
  - `atexit` handler ensures shutdown on normal process exit

[...]

PR #4168, opened 2025-11-07
Add minimal playtime tracking frontend

evertonstz:
Continues work from #4108

Feel free to change things around. My goal was to just make a minimal UI to surface these values and lay the groundwork for future enhancements.

Captura de tela de 2025-11-07 13-15-37

### DIFF
--- a/bottles/backend/managers/playtime.py
+++ b/bottles/backend/managers/playtime.py
@@ -4,6 +4,7 @@
 
 from __future__ import annotations
 
+import re
 import atexit
 import hashlib
 import os
@@ -11,7 +12,7 @@ import sqlite3
 import threading
 import time
 from dataclasses import dataclass
-from typing import Dict, Optional
+from typing import Any, Dict, List, Optional, TypedDict
 
 from bottles.backend.globals import Paths
 from bottles.backend.logger import Logger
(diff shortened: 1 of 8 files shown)


mirkobrombin:
This is exactly how I imagined it!

mirkobrombin (approved):

Perkolator:
Is there any performance penalty using this feature? I mean, for example, how often does it save the values to DB on disk? Could it affect e.g. a game that reads a lot of data from disk?

evertonstz:
> Is there any performance penalty using this feature? I mean, for example, how often does it save the values to DB on disk? Could it affect e.g. a game that reads a lot of data from disk?

I mean, its something extra running, so it will consume resources. That being said, the impact is negligible, I would say: 

- Additional CPU usage: << 0.1% of a single core
- Additional memory: < 10 KB (one session) up to a few hundred KB with many concurrent games running
- Additional disk writes: A handful of small writes per session start/finish plus one write per heartbeat (default to 60 seconds); well below background noise compared to game asset streaming
- Long-term disk space: Likely under a few MB after many years of use for typical play patterns

SQLite is one of the most optimized pieces of software out there, its designed for resource-constrained environments. So long story short: dont worry about it.

Perkolator:
Would it make sense to change the "last played" and "playtime" to "last ran" and "runtime"? I mean, Bottles can be used to run games and programs. Played and playtime feels weird for a program. Ran and runtime are more neutral and work for both programs and games.

evertonstz:
> Would it make sense to change the "last played" and "playtime" to "last ran" and "runtime"? I mean, Bottles can be used to run games and programs. Played and playtime feels weird for a program. Ran and runtime are more neutral and work for both programs and games.

Open an issue asking about it, I'm sure the maintainers will take it in consideration. This PR is already merged and I will probably work on some other projects for a while before incrementing this feat.

PR #4187, opened 2025-11-17
fix[closes #4184]: keep system runner after Wine updates

### LINKED ISSUE #4184: [Bug]: Bottle Switches to Other Runner When sys-wine is Updated
### Describe the bug

Whenever sys-wine is updated, the name of the runner is updated within bottles. When viewing the bottle it will give you an error telling you the runner was not found. If you open a wine app through the bottle it will reconfigure the wine prefix to whatever the fallback runner is. This should not be the default behavior because then wine-sys cannot be reliably set as the runner.

### To Reproduce

1. Wait for your Linux distribution's package manager to update your system wine installation
2. Open a program using the desktop icon
3. The bottle will reconfigure to a fallback runner 

### Package

Flatpak from Flathub

### Distribution

Arch Linux

### Debugging Information

```shell
Official Package: true
Version: '51.24'
DE/WM: plasmawayland
Display:
    X.org: true
    X.org (port): :1
    Wayland: true
Graphics:
     vendors:
        amd:
            vendor: amd
            envs:
                DRI_PRIME: '1'
            icd: /usr/share/vulkan/icd.d/radeon_icd.i686.json:/usr/share/vulkan/icd.d/radeon_icd.x86_64.json
    prime:
        integrated: null
        discrete: null
Kernel:
    Type: Linux
    Version: 6.17.8-arch1-1
Disk:
    Total: 1005922922496
    Free: 400707215360
RAM:
    MemTotal: 15.5GiB
    MemAvailable: 12.2GiB
Bottles_envs: null
```

### Troubleshooting Logs

```shell

```

### Additional context

_No response_

mikayahlevi:
Adds a new runner option named sys-wine-current which will allow the bottle to automatically keep up with the current system version of wine installed. 

# Description
The code just adds the name of that option if a system installation of wine is detected, since any option name starting with "sys-" should just use the system paths/libraries.

Fixes #(4184)

## Type of change

### DIFF
--- a/bottles/backend/managers/manager.py
+++ b/bottles/backend/managers/manager.py
@@ -1283,6 +1283,8 @@ class Manager(metaclass=Singleton):
                 return
 
             config = config_load.data
+            session_arguments = config.session_arguments
+            run_in_terminal = config.run_in_terminal
 
             # Clear Run Executable parameters on new session start
             if config.session_arguments:
@@ -1337,6 +1339,25 @@ class Manager(metaclass=Singleton):
                     value=sample.Parameters[key],
                     scope="Parameters",
                 )
+
+            system_runners = [
+                runner for runner in self.runners_available if runner.startswith("sys-")
+            ]
+            if (
+                config.Environment != "Steam"
+                and config.Runner.startswith("sys-")
+                and len(system_runners) == 1
+                and config.Runner != system_runners[0]
+            ):
+                logging.info(
+                    f"System Wine changed from {config.Runner} to {system_runners[0]}."
+                )
+                config.Runner = system_runners[0]
+                persisted_config = config.copy()
+                persisted_config.session_arguments = session_arguments
+                persisted_config.run_in_terminal = run_in_terminal
+                persisted_config.dump(_config)
+
             self.local_bottles[config.Name] = config
 
             try:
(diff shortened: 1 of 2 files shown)


mirkobrombin:
Hi thanks for your contribution. Are you sure that single line is enough?

mikayahlevi:
Unfortunately my computer is having problems building flatpaks, so I can't test it for sure, but from a code standpoint I am pretty confident it will work. Based on what I saw, Bottles is currently just using janky code for system wine installations. The code only knows whether a runner is a system installation based on whether or not it starts with the prefix 'sys-'. Furthermore, the code to handles wine commands launches wine by setting the PATH environment variable to the location of the runner, but for the system installations it just appends an invalid path and then falls back to the default path (I mention this in [this PR](https://github.com/bottlesdevs/Bottles/pull/4186#issue-3631656987)). I think the code for handling runners should probably also be cleaned up since it's not a very thorough implementation but it should allow me to add the 'sys-wine-current' in one line of code.

https://github.com/bottlesdevs/Bottles/blob/eb96598ff13886dbf78b15f8f153fc0a837e48c0/bottles/backend/wine/wineserver.py#L32C1-L43C52
https://github.com/bottlesdevs/Bottles/blob/eb96598ff13886dbf78b15f8f153fc0a837e48c0/bottles/backend/utils/manager.py#L94C1-L98C43
https://github.com/bottlesdevs/Bottles/blob/eb96598ff13886dbf78b15f8f153fc0a837e48c0/bottles/backend/runner.py#L56C9-L56C42

PR #4196, opened 2025-11-19
feat: ability to spawn processes with (new) Winebridge

mirkobrombin:


Perkolator:
Would be nice to have some help text/page of what Winebridge actually is and what it does (and when to use or not for example).

Coki91:
> Would be nice to have some help text/page of what Winebridge actually is and what it does (and when to use or not for example).

I came here for the same thing, not knowing what winebridge is as all results in the web mostly point to real life wine manufacturing companies, though the organization has it's repository with little explanation of it's functioning https://github.com/bottlesdevs/winebridge

PR #4205, opened 2025-11-20
feat: Re-introduce the journal browser

mirkobrombin:


Perkolator:
Why is it possible to drag and drop the journal entries to different locations in the list?

mirkobrombin:
Just me having a standard day with GTK

PR #4228, opened 2025-11-24
Remove unnecessary runners such as soda

csc-chicken:
# Description

  I found that only kron4ek and proton-ge are actively and automatically built and upload. So, why not just remove others?
  I want to use LINE with wine. I searched on WineHQ and found out it can work under wine 10.18. And that's when I found out every other runners are so outdated. I think this can make newcomer-friendly by leave one that is active.Look! Wine-ge-proton branch is completely archived and stopped at proton 8 . 
  Maybe adding titles that are active or archive is also relevant. 

# Fixes 
Erase runners on preference.py 

# How Has This Been Tested?

I don't know how to build flatpak but look at the codebase I think remove these lines are enough I think.

### DIFF
--- a/bottles/frontend/views/preferences.py
+++ b/bottles/frontend/views/preferences.py
@@ -504,29 +504,11 @@ class PreferencesWindow(Adw.PreferencesWindow):
                 identifiable_runner["count"] += 1
 
     def populate_runners_list(self):
-        exp_soda = ComponentExpander(
-            "Soda", _("Based on Valve's Wine, includes Staging and Proton patches.")
-        )
-        exp_caffe = ComponentExpander(
-            "Caffe", _("Based on Wine upstream, includes Staging and Proton patches.")
-        )
-        exp_wine_ge = ComponentExpander(
-            "Wine GE",
-            _(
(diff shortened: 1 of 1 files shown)


AlmeidaRW:
Should Bottles get rid of wine-ge when umu-launcher support is not implemented? I though GE-Proton sort of needed it, and using it without umu-launcher could cause issues

csc-chicken:
Wine-ge has stopped for a while. So what I did is only leaving kron4ek and proton-ge. Your opinion is important as well

mirkobrombin:
Soda is the default one and work is being done to upgrade it, vanilla and caffe and others are there for compatibility. Can't merge.

csc-chicken:
Yeah! Fair enough! Maybe add title "active" and "archived" is better. BTW, Oh my god! I was replied by the [mirkobrombin](https://github.com/mirkobrombin) yourself ? I am your big fan! Your creation of bottles is phenomenal!

mirkobrombin:
> Your creation of bottles is phenomenal!

<3

I'll evaluate it, thanks for the proposal.

PR #4248, opened 2025-12-02
fix: change space-containing value(s) in Exec key of generated Desktop Entry file…

ChillyVanilly77:
… to be double- instead of single-quoted (in ManagerUtils.create_desktop_entry(…))

Has been tested by locally building the project into the "Devel" flatpak image using the suggested `flatpak-builder` and successfully validated by `desktop-file-validate` util.

Q: any particular reason for using only at 1 place `shlex.quote(program.get('name'))` at (what would have been) line 270?

### DIFF
--- a/bottles/backend/utils/manager.py
+++ b/bottles/backend/utils/manager.py
@@ -263,11 +263,12 @@ class ManagerUtils:
                 for file in existing_files:
                     os.remove(file)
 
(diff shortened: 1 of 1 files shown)


mirkobrombin:
> Q: any particular reason for using only at 1 place `shlex.quote(program.get('name'))` at (what would have been) line 270?

not relly.

PR #4409, opened 2026-02-22
Don't use templates with a mismatched uuid

### LINKED ISSUE #4405: [Bug]: Can't create new bottles
### Describe the bug

Tried creating bottles based on sys-wine-11.0 and soda-9.0-1. Both will stop with error message "Unable to Create Bottle" after "Template found, applying...".

When restarting Bottles, this is shown:

```
[Errno 2] No such file or directory: '/home/USER/.var/app/com.usebottles.bottles/data/bottles/templates/83b27b67-0df5-48af-946d-70cb590db4b3'  File "/app/share/bottles/bottles/backend/utils/threading.py", line 73, in __target
    result = self.task_func(*args, **kwargs)

  File "/app/share/bottles/bottles/backend/managers/manager.py", line 1758, in create_bottle
    TemplateManager.unpack_template(template, config)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^

  File "/app/share/bottles/bottles/backend/managers/template.py", line 191, in unpack_template
    shutil.copytree(
    ~~~~~~~~~~~~~~~^
        _path,
        ^^^^^^
    ...<4 lines>...
        ignore_dangling_symlinks=True,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^

  File "/usr/lib/python3.13/shutil.py", line 591, in copytree
    with os.scandir(src) as itr:
         ~~~~~~~~~~^^^^^

```
True enough, the system does not have a folder named 83b27b67-0df5-48af-946d-70cb590db4b3. It has 2 other folders with similar uuid names. 

Running Bottles v.62.0.
Existing bottles works ok.

### To Reproduce

Create a bottle with type Application and either Wine 11.0 or Soda 9.0-1 runner in the default dir.
Watch it fail to complete the creation.

### Package

Flatpak from Flathub

### Distribution

Pop_OS

### Debugging Information

```shell
Official Package: true
Version: '62.0'
DE/WM: pop
Display:
    X.org: true
    X.org (port): :1
    Wayland: false
Graphics:
    vendors:
        amd:
            vendor: amd
            envs:
                DRI_PRIME: '1'
            icd: /usr/share/vulkan/icd.d/radeon_icd.x86_64.json:/usr/lib/x86_64-linux-gnu/GL/vulkan/icd.d/radeon_icd.x86_64.json:/usr/lib/i386-linux-gnu/GL/vulkan/icd.d/radeon_icd.i686.json
    prime:
        integrated: null
        discrete: null
Kernel:
    Type: Linux
    Version: 6.17.9-76061709-generic
Disk:
    Total: 32429117440
    Free: 32428953600
RAM:
    MemTotal: 60.4GiB
    MemAvailable: 48.9GiB
Bottles_envs: null
```

### Troubleshooting Logs

```shell

```

### Additional context

_No response_

Graham--M:
## Overview
Fixes an issue where incomplete template recreation can leave a template with an older `template.yml` and referencing a non-existent uuid.

Bottles uses the uuid entry in the `template.yml` to build a path to the directory, resulting in  `[Errno 2] No such file or directory` errors when attempting to unpack or delete the template.

Fixes #4405

## Type of change

## Description of Bug
1. When creating a bottle prefix using a built-in environment and a template for it exists, the template is unpacked into the bottle prefix directory including the `template.yml` from the template.
2. One of the components in the new bottle config (e.g. DXVK) can differ from the one in the template causing an update.
3. The old template is deleted and a directory for the new template is created with a new uuid.
4. The bottle prefix contents are copied to the new template directory including the previous `template.yml`. This happens because the ignore patterns contain a typo so no `.yml` files are filtered.
5. Other issues can cause the template creation to be interrupted before the new `template.yml` can be written.
6. Now when attempting to create another new bottle prefix in the same environment, unpacking the template (or removing it if out of date) will fail as the template is not located in a directory with the same uuid as in the `template.yml`.

## How Has This Been Tested?
- Modify an existing template's `template.yml` to list a different DXVK version and create a new bottle prefix in the same environment (or use the CLI tool to create the bottle prefix with a different DXVK version). Inspect the new template folder to see if the `bottle.yml` was filtered out (if it exists then `.yml` files are not being filtered).
- Modify an existing template's `template.yml` to have a different uuid and create a new bottle prefix. It should ignore the old template and a create a new one.

## Other notes
- The modified ignore patterns should prevent this specific issue occurring in the future.
- Verifying that the directory uuid matches the `template.yml` fixes issues caused by existing templates with this problem.
- When the uuids don't match, the broken template is simply ignored in line with how the current code ignores templates without a `template.yml`. However this does require users with broken templates to manually remove them to reclaim storage space. Only valid templates can be deleted from the GUI.

### DIFF
--- a/bottles/backend/managers/template.py
+++ b/bottles/backend/managers/template.py
@@ -124,7 +124,7 @@ class TemplateManager:
         for template in templates:
             if os.path.exists(os.path.join(Paths.templates, template, "template.yml")):
                 _manifest = TemplateManager.get_template_manifest(template)
-                if _manifest is not None:
+                if _manifest is not None and _manifest["uuid"] == template:
                     res.append(_manifest)
 
         return res


mirkobrombin:
LGTM, btw I’d also change _manifest["uuid"] == template to _manifest.get("uuid") == template so a malformed template.yml is ignored instead of raising. But it's a good solution.

Issue #4402, opened 2026-02-17
[Bug]: Unable to create start menu entries


jedikosh:
### Describe the bug

I am unable to create start menu entries. 

Ran: flatpak override com.usebottles.bottles --user --filesystem=xdg-data/applications

Restarted the Bottles. Attempted to add the entry and pressed create. No start menu entry.

### To Reproduce

Unsure how to create error as the process has not worked.

### Package

Flatpak from Flathub

### Distribution

Linux Mint 22.3

### Debugging Information

```shell
Official Package: true
Version: '62.0'
DE/WM: cinnamon
Display:
    X.org: true
    X.org (port): :0
    Wayland: false
Graphics:
    vendors:
        nvidia: &amp;id002
            vendor: nvidia
            envs:
                __NV_PRIME_RENDER_OFFLOAD: '1'
                __GLX_VENDOR_LIBRARY_NAME: nvidia
                __VK_LAYER_NV_optimus: NVIDIA_only
            icd: /usr/share/vulkan/icd.d/nvidia_icd.json
            nvngx_path: /usr/lib/x86_64-linux-gnu/GL/nvidia-580-126-09/lib/nvidia/wine
        intel: &amp;id001
            vendor: intel
            envs:
                DRI_PRIME: '1'
            icd: /usr/share/vulkan/icd.d/intel_hasvk_icd.x86_64.json:/usr/share/vulkan/icd.d/intel_icd.x86_64.json:/usr/lib/x86_64-linux-gnu/GL/vulkan/icd.d/intel_hasvk_icd.x86_64.json:/usr/lib/x86_64-linux-gnu/GL/vulkan/icd.d/intel_icd.x86_64.json:/usr/lib/i386-linux-gnu/GL/vulkan/icd.d/intel_hasvk_icd.i686.json:/usr/lib/i386-linux-gnu/GL/vulkan/icd.d/intel_icd.i686.json
    prime:
        integrated: *id001
        discrete: *id002
Kernel:
    Type: Linux
    Version: 6.17.0-14-generic
Disk:
    Total: 12521463808
    Free: 12521275392
RAM:
    MemTotal: 23.3GiB
    MemAvailable: 18.4GiB
Bottles_envs: null
```

### Troubleshooting Logs

```shell

```

Simon-Spettmann:
[...]

PR #4417, opened 2026-03-01
Fix some ICO files not converting to PNG

ignapk:
# Description

Some ICO files are incorrectly identified as TARGA by libmagic file utility, used to obtain the MIME type of the extracted icon from the executable while creating a desktop entry using dynamic launcher portal. They are then assumed to be already PNG, and therefore are not correctly converted, which results in failure to validate the icon by the portal.

While this is a libmagic bug, workaround it by assuming that "image/x-tga" MIME type also indicates ICO files.

See https://bugs.astron.com/view.php?id=723

Helps #4402

## Type of change

# How Has This Been Tested?
- [ ] Create a desktop entry for NFS most wanted (2012) (NFS13.exe)

### DIFF
--- a/bottles/backend/utils/manager.py
+++ b/bottles/backend/utils/manager.py
@@ -206,7 +206,9 @@ class ManagerUtils:
                 os.remove(ico_dest)
 
             ico.export_icon(ico_dest_temp)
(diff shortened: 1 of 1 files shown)


mirkobrombin:
LGTM

PR #4435, opened 2026-03-16
Drop deprecated vars

ndiruhniu:
# Description
DXVK_STATE_CACHE_PATH is no longer used in the current DXVK, so it has been replaced with DXVK_SHADER_CACHE_PATH.

__GL_SHADER_DISK_CACHE_SKIP_CLEANUP is obsolete in modern Mesa.

__GL_SHADER_DISK_CACHE_PATH is obsolete in modern systems, as it is handled through Mesa via MESA_SHADER_CACHE_DIR

## Type of change

# How Has This Been Tested?
It has not been tested, but nothing should break.

### DIFF
--- a/bottles/backend/wine/winecommand.py
+++ b/bottles/backend/wine/winecommand.py
@@ -366,6 +366,10 @@ class WineCommand:
             env.add("STAGING_SHARED_MEMORY", "1")
             env.add("__GL_SHADER_DISK_CACHE", "1")
             env.add(
+                "__GL_SHADER_DISK_CACHE_PATH",
+                os.path.join(bottle, "cache", "gl_shader"),
+            )
+            env.add(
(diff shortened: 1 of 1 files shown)


mirkobrombin (changes requested):
We cannot simply drop them, Bottles provides older versions too. Both must be supported

ndiruhniu:
This does not break compatibility with previous versions. The removed OpenGL cache flags cannot break anything, and DXVK has been using DXVK_SHADER_CACHE_PATH instead of STATE cache since version 2.7. Even if an old DXVK and Mesa are used, it will not harm the user.

ndiruhniu:
I tested this with some games, and everything works perfectly.

mirkobrombin:
Thanks for the fix.

Issue #4027, opened 2025-08-07
[Bug]: NTSync not working on Proton GE and other Wine enabled version


VinegarLove:
### Describe the bug

All version of NTSync compatible version don't enable it by default, as it should be when ESYNC and FSYNC are disabled. That is, chosing "System" in the sync option.

This doesn't happen launching every Wine version from:
- Steam custom game
- Lutris
- umu-launcher
- Manually via commandline

This bug is present both in Flathub version and Native packaged version.

### To Reproduce

1. Install any Wine enabled NTSync version
2. Install a custom application, possible that uses VK3D or similar to test
3. Enable MangoHud
4. Run the Application, check MangoHud and see that the WSync is neither set to ESYNC, FSYNC or NTSync.

### Package

Flatpak from Flathub

### Distribution

Arch Linux

### Debugging Information

```shell
Official Package: false
Version: '51.21'
DE/WM: gnome
Display:
    X.org: true
    X.org (port): :0
    Wayland: true
Graphics:
    vendors:
        amd:
            vendor: amd
            envs:
                DRI_PRIME: '1'
            icd: /usr/share/vulkan/icd.d/radeon_icd.i686.json:/usr/share/vulkan/icd.d/amd_icd64.json:/usr/share/vulkan/icd.d/radeon_icd.x86_64.json
    prime:
        integrated: null
        discrete: null
Kernel:
    Type: Linux
    Version: 6.15.9-arch1-1
Disk:
    Total: 105089261568
    Free: 40416731136
RAM:
    MemTotal: 30.9GiB
    MemAvailable: 23.0GiB
Bottles_envs: null
```

### Troubleshooting Logs

```shell

```

syntheticbit:
The comment https://github.com/bottlesdevs/Bottles/issues/3905#issuecomment-3160420226 has details. 
> Bottles doesn't use the proton script to set the correct environment variables so at least for now you need to set WINENTSYNC=1.

mongaikan:
> The comment [#3905 (comment)](https://github.com/bottlesdevs/Bottles/issues/3905#issuecomment-3160420226) has details.
> 
> > Bottles doesn't use the proton script to set the correct environment variables so at least for now you need to set WINENTSYNC=1.

Doesn't work for me with runner sys-wine-11.0
```
flatpak run --command=sh com.usebottles.bottles -c 'ls -l /dev/ntsync'
[...]

PR #4489, opened 2026-04-20
feat: add ntsync support

### LINKED ISSUE #3905: [Request]: NTsync wine verstion and ntsync option from drop down menu

NTsync is on the rage right now, i think it would be great if there would be an option to use ntsync in bottles 

Have ntsync tkg version in runners 

### Other solutions?

_No response_

_No response_

Tid0lla:
# Description
Added option to select NTSync as Synchronization driver
image

NTSync driver has been officially supported since Wine 11 & some custom Proton builds. It can provide substantially better performance than Fsync or Esync. As a result, I have also changed default gaming template synchronization driver to it.

Fixes #3905 , #4027

## Type of change

# How Has This Been Tested?
1. Choose any bottle & change synchronization setting
2. Eanble Mangohud with verbose settings so that you can check current `WSYNC` 

- [x] Set `Synchronization` to `NTSync`
Expected `WSYNC`:
image

- [x] Set `Synchronization` to `System`
Expected `WSYNC`:
image

- [x] Set `Synchronization` to `Fsync`
Expected `WSYNC`:
image


mirkobrombin:
Hi, thanks for your contribution but there are a few issues with this simple implementation:
1) changing the default is risky, the ntsync module is mainline only since Linux 6.14 and runners before Wine 11 does not support it, in all those scenarios the bottle (and the runner) will not work, a possible solution might be checking for the module and wine version and degrade to fsync, this will also prohibit the user form enabling it from the GUI in case the module is missing or the runner does not support it
2) the proper name is NTSync not NTsync, just a typo

I'll commit those fixes directly in your branch, so we can get this merged.

PR #4497, opened 2026-05-01
fix: drastically slow startup and UI freezes with many programs/library entries

dyegoaurelio:
# Description
Bottles becomes extremely slow (multiple minutes) and completely unresponsive during startup and program-related operations (e.g. renaming) when a bottle has many installed programs. GNOME reports the window as "not responding" during these periods.

Two issues were found in `Manager.get_programs()`:

1. **Third-party store detection was inside the `.lnk` iteration loop.** The Steam, Epic Games, and Ubisoft Connect store checks (including `SteamManager` construction with filesystem I/O) were incorrectly indented inside the `for program in results:` loop, causing them to run once per `.lnk` file instead of once per call. With ~20 shortcuts, this meant ~20 redundant `SteamManager` instantiations.

2. **`get_programs()` was called repeatedly with no caching.** Each `LibraryEntry` widget calls `get_programs()` independently during initialization, so N library entries meant N full filesystem scans of the same bottle. A programs cache on the `Manager` singleton (keyed by bottle name) now ensures the expensive scan runs only once per bottle. The cache is invalidated when bottles are refreshed (`check_bottles()`) or when programs are modified (`update_config` with `External_Programs` scope).

## Type of change

# How Has This Been Tested?
- [x] Startup with a bottle containing ~20 programs and multiple library entries — previously took 2+ minutes with a completely frozen UI, now completes in seconds
- [x] Renaming a program — previously caused a long UI freeze, now responsive
- [x] Verified programs still appear correctly in library and bottle details views
- [x] Verified cache invalidation: adding/removing/renaming programs refreshes correctly

### DIFF
--- a/bottles/backend/managers/manager.py
+++ b/bottles/backend/managers/manager.py
@@ -98,6 +98,7 @@ class Manager(metaclass=Singleton):
     nvapi_available = []
     latencyflex_available = []
     local_bottles: Dict[str, BottleConfig] = {}
+    _programs_cache: Dict[str, List[dict]] = {}
     supported_runtimes = {}
     supported_winebridge = {}
     supported_wine_runners = {}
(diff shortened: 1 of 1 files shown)


mirkobrombin:
Hi, thanks for your contribution!

The code looks good, I just see a possible issue with the fact it returns the cached list by reference, not a copy, if the caller edits it like a sort for example, breaks the cache for all the others. It's a simple fix, I'll do it on another commit.

Issue #3538, opened 2024-10-19
[Bug]: None of my apps have a icon


Wolfey-foryou:
### Describe the bug

All of the apps i open/use on bottles js have the bottles icon
![image](https://github.com/user-attachments/assets/19805199-ec2c-4cc2-bb06-bd8413c9f805)
![image](https://github.com/user-attachments/assets/68a71829-bb38-49d7-9f5d-ac860cbfe0ff)

### To Reproduce

Just start an app with bottles prob only on gnome tho tried on kde its normal

### Package

Flatpak from Flathub

### Distribution

Fedora 40

### Debugging Information

```shell
Official Package: true
Version: '51.15'
DE/WM: gnome
Display:
    X.org: true
    X.org (port): :0
    Wayland: true
Graphics:
    vendors:
        amd:
            vendor: amd
            envs:
                DRI_PRIME: '1'
            icd: /usr/lib/x86_64-linux-gnu/GL/vulkan/icd.d/radeon_icd.x86_64.json:/usr/lib/i386-linux-gnu/GL/vulkan/icd.d/radeon_icd.i686.json
    prime:
        integrated: null
        discrete: null
Kernel:
    Type: Linux
    Version: 6.11.3-200.fc40.x86_64
Disk:
    Total: 8310431744
    Free: 8310259712
RAM:
    MemTotal: 15.5GiB
    MemAvailable: 10.0GiB
Bottles_envs: null
```

mirkobrombin:
Nothing we can fix here, that’s how works on gnome

Issue #3430, opened 2024-07-12
Issue with Displaying Application and Game Icons in Bottles


RaulKong898:
Dear Bottles Support Team,

I hope this message finds you well. I am writing to bring to your attention an issue I have encountered while using the Bottles application on multiple Linux operating systems.

I have noticed that when running various applications and games through Bottles, the taskbar icons display the Bottles logo instead of the individual icons for each application or game. This behavior makes it difficult to quickly identify the running applications. Ideally, I would like the taskbar to display the specific icon for each application or game, reserving the Bottles logo only for the Bottles management interface.

This issue persists across several Linux distributions that I have tried over time to see which best suits my needs.

Attached is a screenshot for your reference. In this screenshot, you can see that even though I have opened the Epic Games application, the icon displayed in the taskbar is still the Bottles icon:

![Screenshot](sandbox:/mnt/data/Screenshot%20from%202024-07-12%2005-56-26.png)

The same thing happens when the application is visible in the area where I can see open applications. Instead of displaying the specific icon of the respective application, it shows the Bottles icon.

Could you please resolve this issue so that the correct icons for each application are displayed in the taskbar? If this feature is not currently supported, I would like to suggest including it in future updates, as it would significantly improve the user experience.

Thank you for your attention to this matter. I look forward to your response.

XEmpty:
I was trying to change wm_class of created game / app entry but not working, maybe because window class name is always steam_proton or any other runner or it's just because game / app is still start in bottles instance.  Did that work before?

yevklim:
I'm launching my Windows apps through `bottles-cli` and was annoyed that all windows are created with WM_CLASS set to "steam_proton".

I was searching for a solution, and eventually, I stumbled upon [this blog post](https://djcev.com/2023/03/appicons-wmclass-wineproton/) from [Cameron Vanderzanden](https://djcev.com/).

In that blog post, he referred to [this commit from 2019](https://github.com/ValveSoftware/wine/pull/47/commits/20915d0ae47229fb51b441cce8f6fedaa9b3ac68), which implemented the mechanism that allows us to set WM_CLASS by setting environment variable SteamAppId before we launch the app.

This solution is not ideal and does not cover all the use cases, but it does cover most of them. Hope this helps you as much as it helped me.

ultradev-coder:
[...]

Issue #3272, opened 2024-02-08
[Bug]: All apps matched with Bottle desktop entry


nadrolinux:
### Describe the bug

Under a Gnome + Wayland all apps started with Bottle are matched with Bottle desktop entry. This is the same issue like with PWA apps and chromium: https://github.com/flathub/org.chromium.Chromium/issues/216
however with chromium we can simple add '--app-id', but I don't see any similar parameter for bottle-cli. Do you have any ideas how we can solve this issue with current Bottle version? I remember that for non-flatpak Steam Proton we could set 'SteamAppId' for matching apps group, maybe Bottles support something similar?

### To Reproduce

Start app installed via Bottle

### Package

Flatpak from Flathub

### Distribution

Fedora 39 Silverblue

### Debugging Information

```shell
Official Package: true
Version: '51.10'
DE/WM: gnome
Display:
    X.org: true
    X.org (port): :0
    Wayland: true
Graphics:
    vendors:
        amd:
            vendor: amd
            envs:
                DRI_PRIME: '1'
            icd: /usr/lib/x86_64-linux-gnu/GL/vulkan/icd.d/radeon_icd.x86_64.json:/usr/lib/i386-linux-gnu/GL/vulkan/icd.d/radeon_icd.i686.json
    prime:
        integrated: null
        discrete: null
Kernel:
    Type: Linux
    Version: 6.7.3-200.fc39.x86_64
Disk:
    Total: 8339992576
    Free: 8339828736
RAM:
    MemTotal: 15.5GiB
    MemAvailable: 10.4GiB
Bottles_envs: null
```

nadrolinux:
I found that I can set wmclass + rename .desktop file to com.usebottles.bottles.CUSTOMNAME.desktop to match apps to different desktop entries.

Issue #2985, opened 2023-07-23
[Request]: Use Real Program Icon for the Program Window?


Perkolator:
### Tell us the problem or your need

The running program window in the taskbar always show the Bottles icon, not the actual program icon.

![bott](https://github.com/bottlesdevs/Bottles/assets/53913724/66b9f5e6-0e7a-4ed5-8472-862cbb5abbc4)

(Screenshot is from Linux Mint 20.3 Cinnamon & X11)

### Describe the solution you'd like

Would it be possible to use the actual program icons for the windows?

PacBilly:
Yeah, I second this request. If you use the non-Flatpak version of Bottles, it handles this properly and shows program icon, but with the Flatpak it shows the Bottles icon instead. Drives me crazy, but the flatpak version is less buggy in other regards, so it's the lesser of two evils, I guess. But even in the Bottles team's example .desktop entry YouTube video, you can see it's running the wrong icon. It needs to appear is though the app is running natively, as it does in the non-Flatpak version.

M0JXD:
I would love this as well, especially given that the guide/docs heavily encourage the use of the Flatpak.

In the meantime, is there any way to set panel/taskbar icons manually in the Flatpak version? I now have multiple apps using Bottles and running them at the same time gets very confusing very quickly!

Ambient-Impact:
I'd love to see this as well, but it looks like this a security restriction of Flatpak on Wayland; see #flatpak/flatpak#4751, so not sure what Bottles can do here.

ParticleMon:
In the OP, the DE is Cinnamon, which is likely on X11 since Cinnamon's Wayland support is still in Alpha.

But if this Flatpak security restriction is "working as designed, although the design is not ideal" (https://github.com/flatpak/flatpak/issues/4751#issuecomment-1041520941) and it also applies to X11, then might the providing of an AppImage be expedited, or at least a .deb?

jntesteves:
I think this issue might get solved when desktop compositors implement this new (staging) Wayland protocol and Flatpak (and maybe Wine also) starts using it: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/269

Ambient-Impact:
[...]

PR #4500, opened 2026-05-05
fix: lowercase executable name for StartupWMClass

EthanChappel:
# Description
Small fix to change StartupWMClass of new desktop entries to match the WM Class of Wine windows for most runners. Before this, GNOME used the Bottles icon for all programs, even with a desktop entry created. Xfce + Docklike Taskbar plugin used a generic icon. Now they should use the correct program icon.

Some runners I tested (soda, proton-EM, ge-proton) always set WM Class to _steam_proton_ when on X11, so this fix does not affect them in that scenario.

Related issues: #2985 #3272 #3430 #3538

## Type of change

# How Has This Been Tested?
Tested on GNOME 50.1, Cinnamon 6.6, and Xfce 4.20 with the Docklike Taskbar plugin. Mostly used sys-wine-11.0 and kron4ek-wine-11.7-amd64 runners.
## Steps
- Install a program in a bottle. I used [WinJUPOS](https://jupos.org/gh/download.htm), but any should do.
- Add Desktop Entry from the Programs list in Bottles.
- Launch program.

### Before in GNOME:
`StartupWMClass=WinJUPOS.x64`

Before screenshot, with the Wine program having a Bottles icon.

### After:
`StartupWMClass=winjupos.x64.exe`

After screenshot, with the Wine program having its own icon.

### DIFF
--- a/bottles/backend/utils/manager.py
+++ b/bottles/backend/utils/manager.py
@@ -259,7 +259,7 @@ class ManagerUtils:
                     Categories=Application;
                     Comment=Launch {} using Bottles.
                     StartupWMClass={}""".format(
-                        exec, program.get("name"), program.get("name")
+                        exec, program.get("name"), program.get("executable").lower()
                     )
                 )
             except GLib.Error as e:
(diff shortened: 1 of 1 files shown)


mirkobrombin (approved):
Thanks! Looks good.

Issue #4510, opened 2026-05-15
[Bug]: bottles-cli run -p still behaves differently from GUI launch on 63.2


louzt:
### Describe the bug

On Bottles Flatpak 63.2, launching a registered GUI program with `bottles-cli run -b <bottle-name> -p <program-name>` still does not behave like launching the same program from the Bottles GUI.

In my case, if the app is already open and I call the same `bottles-cli run -p` command again, Bottles prints messages like:

```text
(INFO) Forcing offline mode
(INFO) Recovered 1 running sessions -> forced at last_seen
(INFO) Windows path detected. Avoiding validation.
```

and the relaunch can end in Wine RPC/OLE errors such as:

```text
Failed to create an IRpcStubBuffer ... 0x80004002
```

The same application works normally when launched from the Bottles GUI, and a local workaround that bypasses `bottles-cli run -p` and uses the bottle's generated `standalone` launcher plus the configured runner `wine` also behaves more predictably.

This looks like the CLI launch path still has lifecycle/state differences compared to the GUI launch path.

### To Reproduce

1. Install Bottles from Flathub.
2. Create a bottle.
3. Install a GUI application in that bottle and add it as a Bottles program.
4. Launch it once.
5. While the application is still open, run:

```bash
flatpak run --command=bottles-cli com.usebottles.bottles run -b "SmartersTV" -p "IPTVSmartersPro-Estable"
```

6. Run the same command again one or more times while the app is still open.

### Expected behavior

The CLI launch path should behave the same as the GUI launch path for the same registered program.

If the app is already running, a second `bottles-cli run -p` invocation should not trigger recovery/offline-state noise or break the running app session.

### Actual behavior

Repeated `bottles-cli run -p` invocations can print `Forcing offline mode`, `Recovered ... running sessions`, and `Windows path detected. Avoiding validation.`

In some runs this also ends in Wine RPC/OLE errors, and the CLI path behaves worse than either:

- launching from the Bottles GUI
- launching through the bottle's generated `standalone` script with the configured runner

### Package

Flatpak from Flathub

### Distribution

Debian GNU/Linux forky/sid

### Debugging Information

```text
Version: 63.2
Display:
    Wayland: true
DE/WM: niri
```

### Troubleshooting Logs

```text
(INFO) Forcing offline mode
(INFO) Recovered 1 running sessions -> forced at last_seen
(INFO) Windows path detected. Avoiding validation.
[...]

PR #4511, opened 2026-05-16
[Bug]: avoid Manager startup side effects during CLI launches

louzt:
# Description
Related #4510

## Summary
CLI entrypoints currently instantiate `Manager` directly. In the `bottles-cli run` case, that means a short-lived CLI invocation can still trigger `Manager` startup work that is useful for the main app lifecycle, but noisy or disruptive for a one-shot launch.

## Why
While reproducing #4510, two startup behaviors stood out:
- the initial connection check still runs even when the connection helper is already in forced offline mode, which produces misleading offline logging during CLI launches
- playtime session recovery can mark sessions as forced even when they still belong to another live Bottles process

## What changed
- skip the initial `check_connection()` call when `Manager` is already in forced offline mode
- skip `recover_open_sessions()` for CLI `Manager` instances
- add the missing `get_int()` fallback to `GSettingsStub`
- harden the affected regression tests around singleton state and CLI session recovery

## How Has This Been Tested?
- [x] `pytest bottles/tests/backend/manager/test_manager.py bottles/tests/backend/integration/playtime/test_recovery.py -q`

## Design note
I kept this change in `Manager` because CLI entrypoints already instantiate `Manager` directly today, and I wanted the smallest safe fix for #4510 without widening bootstrapping changes in the same PR.

## Risks / Follow-ups
- Scoped to `Manager` startup behavior for CLI launches and forced-offline initialization.
- Does not change executor-side Windows path validation warnings.

### DIFF
--- a/bottles/backend/managers/manager.py
+++ b/bottles/backend/managers/manager.py
@@ -130,7 +130,7 @@ class Manager(metaclass=Singleton):
         self.data_mgr = DataManager()
         _offline = True
 
-        if check_connection:
+        if check_connection and not self.is_cli:
             _offline = not self.utils_conn.check_connection()
 
         # validating user-defined Paths.bottles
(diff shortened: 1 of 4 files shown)


mirkobrombin:
Hi, thanks for your MR!

Changes looks good but the RPC error mentioned in the issue should have its dedicated issue.

PR #4532, opened 2026-05-30
Add wayland option in launch options

Fredolx:
# Description
This PR adds a wayland option per-executable in the launch settings.
It is useful if the global wayland settings is useful for most executables in a bottle, except a few (white screen, crash)

## Type of change

# How Has This Been Tested?
- Create a new bottle
- Enable wayland globaly on the bottle
- Add an executable with "Add shortcuts"
- In launch option, toggle off the new wayland option.
- Launch the executable

### DIFF
--- a/bottles/backend/managers/manager.py
+++ b/bottles/backend/managers/manager.py
@@ -1127,6 +1127,7 @@ class Manager(metaclass=Singleton):
                     "pulseaudio_latency": _program.get("pulseaudio_latency"),
                     "virtual_desktop": _program.get("virtual_desktop"),
                     "winebridge": _program.get("winebridge"),
+                    "wayland": _program.get("wayland"),
                     "removed": _program.get("removed"),
                     "id": _program.get("id"),
                 }
--- a/bottles/backend/wine/executor.py
+++ b/bottles/backend/wine/executor.py
@@ -60,6 +60,7 @@ class WineExecutor:
         program_gamescope: Optional[bool] = None,
         program_virt_desktop: Optional[bool] = None,
         program_winebridge: Optional[bool] = None,
+        program_wayland: Optional[bool] = None,
         sandbox_override: Optional[str] = None,
     ):
         logging.info("Launching an executable…")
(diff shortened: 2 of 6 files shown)


mirkobrombin:
I was not aware the `WAYLAND` var env was a thing D: Not sure how to test it since I already use wayland only, I'll wait for a second opinion. Thanks!

louzt:
### Architectural Review & Second Opinion

This PR provides a highly necessary feature for per-executable runtime isolation. In complex winebottles, a global Wayland configuration might work flawlessly for 90% of the applications, while specific legacy executables or media players suffer from rendering glitches (e.g., zero-frame white screens or window lifecycle crashes). Myself had this problem on SmartersIPTV app time ago, sometimes worked, sometimes clics didnt work, sometimes white screen, most of times works but... enough for me and reason why i change my client towards open-tv/fredtv months ago. 

Allowing a granular, per-executable override to safely decouple specific processes back to XWayland/X11 without mutating the entire bottle layer is the correct architectural approach in my op.

#### Clarification on the `WAYLAND` Environment Variable:
To clear up any confusion regarding the execution block: `WAYLAND` is **not** being introduced here as an upstream Linux OS or Wine native environment variable. It is being leveraged strictly as an **internal transport token** inside Bottles' internal `environment` dictionary pipeline to safely pipe the evaluation state from `executor.py` down to `winecommand.py` without breaking pre-existing serialization signatures. 

Even though it gets exported to the spawned child process environment block, it remains completely inert to the host OS (which relies entirely on `WAYLAND_DISPLAY`) but flawlessly fulfills its role as an internal state conduit.

#### Resolution Matrix Workflow:

```mermaid
flowchart TD
    A["Launch Executable Triggered"] --> B{"Is per-executable wayland explicit?"}
    B -->|Yes| C{"Does it differ from global?"}
    B -->|No| F["Fallback to Global Parameter"]
    
    C -->|Yes - Override| D["Inject internal flag: WAYLAND = 1 or 0"]
    C -->|No - Redundant| F
    
    D --> E["winecommand.py reads flag from environment"]
    F --> G["winecommand.py reads global via getattr"]
    
    E --> H["Execute: apply_wayland_preferences"]
    G --> H

mirkobrombin:
The per-program override is incomplete in both directions. Enabling it does not enable the Wine Wayland driver in the prefix, while disabling it leaves the inherited WAYLAND_DISPLAY available. Driver selection cannot be safely changed per launch with this environment token, so this approach is not planned.

PR #4560, opened 2026-06-27
feat: detect Unreal Engine bootstrapper in Eagle and suggest the shipping binary

fl-f:
# Description
Unreal Engine games ship with a small launcher whose sole job is to check for prerequisites, before running the real game binary found in Game/Binaries/Win64/Game-Win64-Shipping.exe

This check fails in Wine even when the correct runtime is installed via Bottles' dependency manager, producing a misleading error (pictured). This change makes Eagle recognize this pattern when analyzing Unreal Engine games and suggests to add the shipping binary instead.
Screenshot From 2026-06-27 01-01-10
(typical error when running the Unreal launcher)

Screenshot From 2026-06-27 00-41-16
(how the new feature appears in the Eagle analysis)

## Type of change

### DIFF
--- a/bottles/backend/managers/eagle.py
+++ b/bottles/backend/managers/eagle.py
@@ -634,6 +634,7 @@ class EagleManager:
 
             self._send_step("Scanning directory structure...")
 
+            ue_shipping_exe = None
             exe_base = basename.rsplit(".", 1)[0]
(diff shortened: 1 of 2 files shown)


mirkobrombin:
Thanks!

